PR #538, opened 2014-06-09
add new feature: face center crop

HTChang:
**New Feature**

Picasso is built-in with two types of cropping:  `centerInside()` and `centerCrop()`.  They work well for most images, but sometimes they can cut out faces that are not located in the center of images, producing undesirable thumbnails.

With the help of the native [`FaceDetector`](http://developer.android.com/reference/android/media/FaceDetector.html) in the Android framework, this pull request adds a new API to do what I call a `faceCenterCrop()`.  It crops an image so that faces detected in the image are kept intact as much as possible.  Like `centerCrop()` and `centerInside()`, the `resize()` API must be called together with `faceCenterCrop()`.  For instance,

``` java
Picasso.with(context)
       .load(url)
       .resize(targetWidth, targetHeight)
       .faceCenterCrop()
       .into(imageView)
```

However, the native `FaceDetector` API provided by the Android framework is slow.  You may provide your own implementation of the `FaceDetector`:in Picasso builder

``` java
Picasso.Builder(context)
       .memoryCache(cache)
       .faceDetector(new MyFaceDetector())
       .build();
```

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso/Request.java
@@ -53,13 +53,19 @@ public final class Request {
   /**
    * True if the final image should use the 'centerCrop' scale technique.
    * <p>
-   * This is mutually exclusive with {@link #centerInside}.
+   * This is mutually exclusive with {@link #centerInside} and {@link #faceCenterCrop}.
    */
   public final boolean centerCrop;
   /**
+   * True if the final image should use the 'faceCenterCrop' scale technique.
+   * <p>
+   * This is mutually exclusive with {@link #centerCrop} and {@link #centerInside}.
+   */
+  public final boolean faceCenterCrop;
+  /**
    * True if the final image should use the 'centerInside' scale technique.
    * <p>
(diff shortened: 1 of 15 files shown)


picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -293,13 +310,43 @@ public Builder clearCenterCrop() {
       return this;
     }
 
+    /** Set the face detector for face center crop. */
+    public Builder setFaceDetector(FaceDetector faceDetector) {
```
meisterfuu: @HTChang I have testet this. What other face detectors work with Android?

meisterfuu: I guess it is interesting to know that the Android face detector doesnt always produce the same result with the same sources. Sometime it detects a face, sometimes it detects no face on the image.

herriojr:
was this abandoned?

AizazAZ:
I am using thumbor and thus putting all load of face/feature detection on server, it uses openCV

Issue #581, opened 2014-07-09
Add source `ImageView` parameter for shared `Callback`


stephanenicolas:
Following from #59 , if devs use a shared `Callback` instance on multiple `ImageView`s, it would be useful to get the source `ImageView` when a callback is received : 

``` java
public interface Callback {
  void onSuccess(ImageView source);

  void onError(ImageView source);

  public static class EmptyCallback implements Callback {

    @Override public void onSuccess(ImageView source) {
    }

    @Override public void onError(ImageView source) {
    }
  }
}
```

PR #582, opened 2014-07-09
Solves #581 : Add source ImageView parameter for shared Callback

stephanenicolas:
An additional test has been added to check the ability of Picasso to distinguish sources in case of a shared callback (on an error).

### DIFF
--- a/picasso-sample/src/main/java/com/example/picasso/SampleGalleryActivity.java
+++ b/picasso-sample/src/main/java/com/example/picasso/SampleGalleryActivity.java
@@ -72,7 +72,7 @@ public class SampleGalleryActivity extends PicassoSampleActivity {
     animator.setDisplayedChild(1);
 
     Picasso.with(this).load(image).into(imageView, new EmptyCallback() {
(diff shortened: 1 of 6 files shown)


picasso-sample/src/main/java/com/example/picasso/SampleGalleryActivity.java
```diff
@@ -72,7 +72,7 @@ private void loadImage() {
     animator.setDisplayedChild(1);
 
     Picasso.with(this).load(image).into(imageView, new EmptyCallback() {
-      @Override public void onSuccess() {
+      @Override public void onSuccess(ImageView source) {
```
dnkoutso: Why not call this `target` instead of `source`?

stephanenicolas: It's the source of the event, which appears to be a target in picasso.

JakeWharton:
Callback is used for multiple targets now so we can't just pass in ImageView anymore.

PR #695, opened 2014-10-09
Add density support.

dagezi:
Add a feature to adjust image density to fit device's one. 
For sources which don't supply original density information, Request now has 'density' parameters so that each RequestHandler can refer.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/AssetRequestHandler.java
+++ b/picasso/src/main/java/com/squareup/picasso/AssetRequestHandler.java
@@ -32,8 +32,10 @@ class AssetRequestHandler extends RequestHandler {
       (SCHEME_FILE + ":///" + ANDROID_ASSET + "/").length();
 
   private final AssetManager assetManager;
+  private final Context context;
 
   public AssetRequestHandler(Context context) {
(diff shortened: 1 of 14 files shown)


hayksaakian:
I also need this feature. In my code i show thumbnails in a sidebar.

            Picasso.with(context)
                    .load(metadata.image_url)
                    .resize(180, 0)
                    .into(holder.image);


This produces thumbnails that are sized inconsistently. While they use the same number of pixels, since the size of dots is different on each platform, the physical size of the image is larger on the N7 (because it's lower DPI) than the N5.

Nexus 5 | Nexus 7
------------ | -------------
Nexus 5 | Nexus 7

JakeWharton:
@hayksaakian Use `resizeDimen` with a dimen or scale the value based on density yourself.

lorenzos:
For future readers, here's my solution without changes to the library:

```java
final Transformation adjustDensity = new Transformation() {

	@Override public Bitmap transform(final Bitmap source) {
		source.setDensity(160);
		return source;
	}

	@Override public String key() {
		return "density";
	}
		
};

Picasso.get().load(url).transform(adjustDensity).into(imageView)
```

Issue #691, opened 2014-10-06
Rotate if necessary


julianonunes:
How can I rotate the picture automatically if the photo has been taken in a different orientation?

afollestad:
Not sure why Picasso still can't respect exif rotation when Glide can.

JakeWharton:
Because we don't implement JPEG parsing.

kaiifalcutela:
@JakeWharton so this happens on jpg images only?

tpfaff:
Are there any plans to support parsing the exif tags for orientation?

JakeWharton:
Not unless Android adds an API for it.

afollestad:
@tpfaff checkout [Glide](https://github.com/bumptech/glide), it's used at Target and Google (e.g. Google Photos)

JakeWharton:
You can leave Target off that list.

tpfaff:
@JakeWharton Looks like they added one in support lib 25.1.0 

> 
Added ExifInterface support library. This library unbundles support for reading Exif information from JPEG and raw formatted files and setting the Exif information on JPEG image files.

xanscale:
https://stackoverflow.com/a/66588700/2642478

Issue #689, opened 2014-10-06
Fit horizontally and crop vertically


julianonunes:
How can I make a photo fit horizontally in a dyncamically created ImageView and crop it vertically?

The same question with a sample of my code has been posted on StackOverflow: http://stackoverflow.com/questions/26208266/set-maximum-width-to-load-images-with-picasso/26208302?noredirect=1#comment41099837_26208302

PR #699, opened 2014-10-11
Fixes sizing for EXIF-rotated images

jacobtabak:
See #689 and #691.  Unfortunately those issues don't do a great job of concisely describing the issue.  

Please note this does not address the issue of sizing images with custom rotation with , for example - `requestCreator.rotate(degrees, pivotX, pivotY)`, because the math is not trivial and there should probably be some discussion about the actual desired behavior.

However, I feel like it's still appropriate to make this change because ideally EXIF rotation should be transparent to the developer, whereas `requestCreator.rotate()` would have to be specifically requested.

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
+++ b/picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
@@ -112,7 +112,7 @@ public class BitmapHunterTest {
     verify(dispatcher).dispatchFailed(hunter);
   }
 
-  @Test public void responseExcpetionDispatchFailed() throws Exception {
+  @Test public void responseExceptionDispatchFailed() throws Exception {
     Action action = mockAction(URI_KEY_1, URI_1);
     BitmapHunter hunter = new TestableBitmapHunter(picasso, dispatcher, cache, stats, action, null,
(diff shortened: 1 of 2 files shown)


picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -405,6 +405,18 @@
     assertThat(shadowMatrix.getPreOperations()).containsOnly("rotate 90.0");
   }
 
+  @Test public void exifRotationSizing() throws Exception {
+    Request data = new Request.Builder(URI_1).rotate(-45).resize(5, 10).build();
```
JakeWharton: The `rotate` call here seems useless, right?

jacobtabak: If I recall correctly the rotation value here affects the shadow matrix pre-operations that are tested in line 417.

JakeWharton: Copy/paste, it seems.

JakeWharton:
I'm going to look at this over the weekend. This is spiraling into craziness and we need to get a lid on it.

PR #707, opened 2014-10-22
Material transition

mattprecious:
Based on suggestions here: http://www.google.com/design/spec/patterns/imagery-treatment.html

Give it a whirl and let me know what needs tweaking.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
@@ -19,6 +19,8 @@ import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
 import android.graphics.ColorFilter;
+import android.graphics.ColorMatrix;
+import android.graphics.ColorMatrixColorFilter;
 import android.graphics.Paint;
 import android.graphics.Path;
 import android.graphics.Point;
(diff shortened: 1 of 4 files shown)


picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
```diff
@@ -109,10 +121,24 @@ static void setPlaceholder(ImageView target, int placeholderResId, Drawable plac
...
-        super.draw(canvas);
-        super.setAlpha(alpha);
+        if (useColorMatrixTransition) {
+          colorMatrix.reset();
+          TRANSITION.apply(colorMatrix, fraction);
+
+          // Scale the transition alpha to match the defined alpha on the drawable.
+          colorMatrix.getArray()[ROW_ALPHA + COLUMN_ALPHA] *= (alpha / 255);
+
+          super.setColorFilter(new ColorMatrixColorFilter(colorMatrix));
```
goncalossilva: I would suggest supporting this on L only and using `ColorMatrixColorFilter#setColorMatrix` via reflection to avoid allocations. On pre-L falling back to fade-in seems reasonable.

I'm aware this is not pretty, but currently I don't see other way of avoiding 13 `ColorMatrixColorFilter` for each image that gets faded in.

JakeWharton:
Needs rebase :trollface:

mattprecious:
Back away slowly...

JakeWharton:
I still want to enable this somehow. Maybe we can resurrect it in the
sample eventually.

Issue #769, opened 2014-11-27
fit().centercrop() doesn't match the width of imageview


raylee4204:
when using fit().centercrop() on an imageview with `match_parent` width, the resized image doesn't fill the width of the imageview.

The following screenshot was taken from a Nexus 9 with lollipop.

![device-2014-11-27-151750](https://cloud.githubusercontent.com/assets/2085389/5221457/dafdbebe-7648-11e4-8071-4104d312b54a.png)

JakeWharton:
Fix was merged. Will be in next release.

PR #784, opened 2014-12-09
Possible fix for #769

iamstuffed:
When Center Crop is used, a single scale factor is calculated to transform the input image to a target image size. This does not work for all input cases. I'm not sure if this is the best solution but hope this at least starts a discussion. The summary of the problem: integers multiplied by a transformation matrix of floats cannot always create the requested targetSize. A single scale factor cannot be used for both X and Y because the input sizes are integers. Test included.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -451,19 +451,21 @@ class BitmapHunter implements Runnable {
       if (data.centerCrop) {
         float widthRatio = targetWidth / (float) inWidth;
         float heightRatio = targetHeight / (float) inHeight;
-        float scale;
+        float scaleX, scaleY;
         if (widthRatio > heightRatio) {
-          scale = widthRatio;
           int newSize = (int) Math.ceil(inHeight * (heightRatio / widthRatio));
           drawY = (inHeight - newSize) / 2;
           drawHeight = newSize;
+          scaleX = widthRatio;
+          scaleY = targetHeight / (float) drawHeight;
         } else {
-          scale = heightRatio;
           int newSize = (int) Math.ceil(inWidth * (widthRatio / heightRatio));
           drawX = (inWidth - newSize) / 2;
           drawWidth = newSize;
+          scaleX = targetWidth / (float) drawWidth;
+          scaleY = heightRatio;
         }
-        matrix.preScale(scale, scale);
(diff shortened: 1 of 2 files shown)


iamstuffed:
Ping.

JakeWharton:
We'll get this in the next release. Don't you worry :)

picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
```diff
@@ -446,19 +446,29 @@ static Bitmap transformResult(Request data, Bitmap result, int exifRotation) {
...
+        float scaleX, scaleY;
         if (widthRatio > heightRatio) {
-          scale = widthRatio;
+          scaleX = widthRatio;
+          scaleY = widthRatio;
           int newSize = (int) Math.ceil(inHeight * (heightRatio / widthRatio));
           drawY = (inHeight - newSize) / 2;
           drawHeight = newSize;
+          int rounded = Math.round(drawHeight * scaleY);
+          if (rounded != targetHeight) {
```
JakeWharton: Is it worth it to do the round and this comparison?

Why not just always set the `scaleY` to `targetHeight / (float) newSize` (and similar below)?

iamstuffed: It's probably okay just to set it. I don't have a good reason for doing it this way. I just didn't want to introduce any additional rounding errors.

I'll update it.

JakeWharton: Great. Let me know.

iamstuffed:
[INFO] Reactor Summary:
[INFO] 
[INFO] Picasso (Parent) .................................. SUCCESS [0.253s]
[INFO] Picasso ........................................... SUCCESS [12.935s]
[INFO] Picasso Pollexor Transformer ...................... SUCCESS [2.988s]
[INFO] Picasso Sample .................................... SUCCESS [7.295s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 24.499s
[INFO] Finished at: Wed Jan 07 22:30:44 PST 2015
[INFO] Final Memory: 31M/361M
[INFO] ------------------------------------------------------------------------

JakeWharton:
Beautiful! Thanks a lot for doing this work.

iamstuffed:
Thanks!

PR #806, opened 2014-12-22
Add stableKey parameter to Downloader.load(...)

meisterfuu:
This little change gives the user the possibility to implement his own
disc chaching with stableKeys within a custom Downloader.

I down't know if is the best way to do this, but is a way without touching picasso very much. One could achieve the same with a custom Cache, but you wouldn't be able to report correct value for Picasso.LoadedFrom without changing the BitmapHunter.

I don't mind if this does not make it in if you feel this is unnecessary but i think it is a logical step after my last PR wich added StableKeys. This code should have no effect on Picasso core thus the null values in the test classes.

I just wanted to share this piece of code, so that maybe someone else can use this. I know this solution is not the best, but it suits exactly my needs.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Downloader.java
+++ b/picasso/src/main/java/com/squareup/picasso/Downloader.java
@@ -28,12 +28,14 @@ public interface Downloader {
    * @param uri Remote image URL.
    * @param localCacheOnly If {@code true} the URL should only be loaded if available in a local
    * disk cache.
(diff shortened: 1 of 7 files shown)


JakeWharton:
The `Downloader` abstraction has been removed on `master` and has been replaced with an interceptor pattern. This allows you to write one which delegates to have the network handler download the image and then write the result to disk using the stable key that's in the request.

PR #821, opened 2014-12-31
Introduce Network Policy API

### LINKED ISSUE #438: Declarative Cache Control API
I've been sitting on this for a while as a future feature but hadn't documented it.

The API would allow specifying a cache interaction policy both "downward" on the load and "upward" back through the processing. Through this you can facilitate complex cache interactions for both valid and seemingly-questionable use cases.

Random subset of interactions from other issues:
- Skip memory cache "downward" only.
- Skip memory cache "upward" only.
- Load from local cache (memory or disk) only.
- Force re-load from network.

I'm not going to do a full-on API strawman, but I know we'll probably want some `EnumSet` action with

``` java
public enum BitmapSource { MEMORY, DISK, NETWORK }
public enum BitmapSink { MEMORY, DISK }
```

This API will trump the existing weak `skipCache()` method on `RequestCreator`.

dnkoutso:
closes #438

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Downloader.java
+++ b/picasso/src/main/java/com/squareup/picasso/Downloader.java
@@ -26,14 +26,13 @@ public interface Downloader {
    * Download the specified image {@code url} from the internet.
    *
    * @param uri Remote image URL.
-   * @param localCacheOnly If {@code true} the URL should only be loaded if available in a local
-   * disk cache.
+   * @param networkPolicy The {@link NetworkPolicy} used for this request.
    * @return {@link Response} containing either a {@link Bitmap} representation of the request or an
    * {@link InputStream} for the image data. {@code null} can be returned to indicate a problem
    * loading the bitmap.
    * @throws IOException if the requested URL cannot successfully be loaded.
    */
-  Response load(Uri uri, boolean localCacheOnly) throws IOException;
+  Response load(Uri uri, int networkPolicy) throws IOException;
 
   /**
    * Allows to perform a clean up for this {@link Downloader} including closing the disk cache and
@@ -46,9 +45,9 @@ public interface Downloader {
     final boolean localCacheOnly;
     final int responseCode;
 
-    public ResponseException(String message, boolean localCacheOnly, int responseCode) {
+    public ResponseException(String message, int networkPolicy, int responseCode) {
       super(message);
-      this.localCacheOnly = localCacheOnly;
+      this.localCacheOnly = NetworkPolicy.isOfflineOnly(networkPolicy);
       this.responseCode = responseCode;
     }
   }
--- a/picasso/src/main/java/com/squareup/picasso/NetworkPolicy.java
+++ b/picasso/src/main/java/com/squareup/picasso/NetworkPolicy.java
@@ -1,0 +1,36 @@
+package com.squareup.picasso;
+
+/** Designates the policy to use for network requests. */
+@SuppressWarnings("PointlessBitwiseExpression")
+public enum NetworkPolicy {
+
+  /** Skips checking the disk cache and forces loading through the network. */
+  NO_CACHE(1 << 0),
+
+  /** Skips storing the result into the disk cache.
+   * <p>
+   * <em>Note</em>: At this time this is only supported if you are using OkHttp.
+   */
+  NO_STORE(1 << 1),
+
+  /** Forces the request through the disk cache only, skipping network. */
+  OFFLINE(1 << 2);
+
+  public static boolean shouldReadFromDiskCache(int networkPolicy) {
+    return (networkPolicy & NetworkPolicy.NO_CACHE.index) == 0;
+  }
+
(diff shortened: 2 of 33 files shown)


dnkoutso:
Updated. Still need to add a few tests.

picasso/src/main/java/com/squareup/picasso/Downloader.java
```diff
@@ -26,14 +26,13 @@
...
-   * @param localCacheOnly If {@code true} the URL should only be loaded if available in a local
-   * disk cache.
+   * @param networkPolicy The {@link NetworkPolicy} used for this request.
    * @return {@link Response} containing either a {@link Bitmap} representation of the request or an
    * {@link InputStream} for the image data. {@code null} can be returned to indicate a problem
    * loading the bitmap.
    * @throws IOException if the requested URL cannot successfully be loaded.
    */
-  Response load(Uri uri, boolean localCacheOnly) throws IOException;
+  Response load(Uri uri, int networkPolicy) throws IOException;
```
dnkoutso: Added it now.

picasso/src/main/java/com/squareup/picasso/OkHttpDownloader.java
```diff
@@ -89,19 +90,29 @@ protected final OkHttpClient getClient() {
...
+        builder.noCache();
+      }
+      if (!NetworkPolicy.shouldWriteToDiskCache(networkPolicy)) {
+        builder.noStore();
+      }
+      cacheControl = builder.build();
     }
 
+    com.squareup.okhttp.Request.Builder requestBuilder =
+        new com.squareup.okhttp.Request.Builder().cacheControl(cacheControl).url(uri.toString());
```
dnkoutso: Actually I am keeping the builder and adding a `networkPolicy > 0` conditional to avoid allocating an extra cache control object.

dnkoutso:
Added some tests. Should be ready now.

picasso/src/main/java/com/squareup/picasso/RequestCreator.java
```diff
@@ -487,7 +511,7 @@ public void into(RemoteViews remoteViews, int viewId, int notificationId,
 
     RemoteViewsAction action =
         new NotificationAction(picasso, request, remoteViews, viewId, notificationId, notification,
-            memoryPolicy, errorResId, key, tag);
+            memoryPolicy, networkPolicy, key, tag, errorResId);
 
     performRemoteViewInto(action);
```
dimlek: k

picasso/src/main/java/com/squareup/picasso/OkHttpDownloader.java
```diff
@@ -89,19 +91,33 @@ protected final OkHttpClient getClient() {
     return client;
   }
 
-  @Override public Response load(Uri uri, boolean localCacheOnly) throws IOException {
-    com.squareup.okhttp.Request.Builder requestBuilder =
-        new com.squareup.okhttp.Request.Builder().url(uri.toString());
+  @Override public Response load(Uri uri, int networkPolicy) throws IOException {
+    CacheControl cacheControl = null;
+    if (networkPolicy > 0) {
```
JakeWharton: This should be `!= 0` since it's a bit field

JakeWharton:
Rebase

picasso/src/main/java/com/squareup/picasso/NetworkPolicy.java
```diff
@@ -0,0 +1,36 @@
+package com.squareup.picasso;
+
+/** Designates the policy to use for network requests. */
+@SuppressWarnings("PointlessBitwiseExpression")
+public enum NetworkPolicy {
+
+  /** Skips checking the disk cache and forces loading through the network. */
+  NO_CACHE(1 << 0),
+
+  /** Skips storing the result into the disk cache.
```
JakeWharton: nit: this should be on the next line

picasso/src/main/java/com/squareup/picasso/Downloader.java
```diff
@@ -46,9 +45,9 @@
     final boolean localCacheOnly;
     final int responseCode;
 
-    public ResponseException(String message, boolean localCacheOnly, int responseCode) {
+    public ResponseException(String message, int networkPolicy, int responseCode) {
```
daisy1754: Sorry for late comment, but I actually have two questions:
1. Why it's exceptional? It seems changing existing public method signature.
2. Why it's _networkPolicy_ is _int_ while _NetworkPolicy_ is defined as _enum_ ?

dnkoutso: For 2).

enum helps shape the API by disallowing any int being passed in. int is used internally to preserve object allocation by performing bitwise operations instead.

daisy1754: Thanks for prompt answer!
By API you mean RequestCreator#networkPolicy? Isn't Downloader public API as well?

PR #834, opened 2015-01-05
Fix nit, add copyright

dnkoutso:


picasso/src/main/java/com/squareup/picasso/NetworkPolicy.java
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
JakeWharton: OMG IT'S THE FUTURE!

Issue #836, opened 2015-01-06
 I get half photo and half white color, and sometimes I get only the white color.


Kuoyhout:
I try to load photo in to girdview sometimes a few photos show only half photo and white color for other half, and some the photos show only white color just like the screenshot below.

![15 - 2](https://cloud.githubusercontent.com/assets/5210467/5628737/e519fade-95dd-11e4-8c70-e8369eb36ad9.jpg)
![15 - 1](https://cloud.githubusercontent.com/assets/5210467/5628742/ec230762-95dd-11e4-8426-ea48d180cb4c.jpg)

JakeWharton:
What version of Picasso? Does it always happen with the same url? What shows up in the logs? Can you reproduce it reliably?

I suspect this has nothing to do with Picasso and it Android's bitmap decoder.

dnkoutso:
You can also try another image loading library and see if this issue occurs.

If it does, can you please provide a sample app so we can fix in Picasso as well?

Kuoyhout:
@JakeWharton I'm using version 2.4.0. It doesn't happen with the same url but when the photo show only half or show only white color, it will stay that way when the photo reload from cache. It doesn't usually happen, sometimes it's working fine, sometimes it give me this issues. I forget to copy the logs when this happen.

@dnkoutso I only know how to use Picasso and Universal image loader. I used to use universal image loader, I didn't get this issue, but I got another issue so I switched to Picasso.

6a209:
I also encountered this problem

jacobtabak:
I am intermittently experiencing the same issue.  Source photo had just been taken with the Nexus 6 camera on Android 5.0.1 and had been loaded with FileRequestHandler.

![screenshot](http://i.imgur.com/2opRDdX.jpg)

jacobtabak:
@Kuoyhout what device are you experiencing this on?  We are seeing it happen exclusively on Nexus 6.

Kuoyhout:
@jacobtabak I experience this on my Galaxy Note 4

dnkoutso:
If it works on another image loading library I'd like for us to tackle this for Picasso. @Kuoyhout got a sample app?

Kuoyhout:
@dnkoutso sorry, I don't have any sample app. @jacobtabak Do you have any sample app?

jacobtabak:
[...]

PR #844, opened 2015-01-12
fixed issues #836

6a209:


JakeWharton:
We are not going to buffer every download. This will destroy the heap.

Issue #427, opened 2014-03-01
Placeholder is stretched before updating the ImageView (2.2.0)


kaciula:
In version 2.2.0 the placeholder image is stretched right before being replaced with the actual image. It looks ugly. In previous version 2.1.1, this worked as expected. Here is the relevant code:

mPicasso.load(url).placeholder(R.drawable.bird_placeholder).error(R.drawable
                .bird_placeholder).into(holder.photo);

holder.photo is a SquaredImageView, meaning it has the same width and height. And the scaleType is "centerCrop".

dlew:
I've noticed that this still seems to be happening in the latest Picasso.

If the ImageView is `centerInside`, but the request uses `fit().centerCrop()`, then at the last second (before it fully loads) the placeholder expands as if the scaletype were actually `centerCrop`.

JakeWharton:
Yes. `PicassoDrawable` doesn't honor the `scaleType` for the placeholder.

dlew:
I guess I'm a bit confused, then, why it *changes* size. If it doesn't honor `scaleType` the entire time, why does it initially start at the correct size, then resize for a split second before the image loads fully?

dlew:
Discovered a bit more info: if I use `noFade()` then the stretching doesn't occur. So the resizing only happens during the fade.

JakeWharton:
Ah, sorry for not elaborating. We set the placeholder directly on the `ImageView` initially. When the URL (or whatever) has finished loading, we copy the existing drawable into a `PicassoDrawable` instance and then do a ~200ms cross-fade between the two. Since `PicassoDrawable` doesn't honor `scaleType`, if the image is being resized by the `ImageView` you'll see it jump to a different size during the cross fade.

The fix would be to copy `ImageView`'s math for `scaleType` and apply it to the placeholder drawable that we steal from it.

dlew:
Ah, that falls in line with something else I just discovered, which was that using `noFade()` fixes the problem (at the cost of removing the fade).

I might look into doing a PR to do what you propose; seems like it should be possible in any case where it's being loaded into an `ImageView`.

dlew:
@JakeWharton I've been investigating the issue and duplicating `ImageView`'s logic is going to be tough, since it would depend heavily on the `ImageView` itself being accessible in `PicassoDrawable` (which I'm guessing is not what we want).

[...]

PR #855, opened 2015-01-19
Use TransitionDrawable for PicassoDrawable

dlew:
Before, it was using a BitmapDrawable with a hacked-in placeholder.
This often caused the placeholder to draw with the incorrect aspect
ratio during loading.

Using a LayerDrawable fixes the aspect ratio; using TransitionDrawable
gives us the facility for cross-fading built into Android.

This PR addresses #427 (though we should determine if it does so in a satisfactory manner).

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
@@ -18,25 +18,24 @@ package com.squareup.picasso;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
-import android.graphics.ColorFilter;
+import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.Path;
 import android.graphics.Point;
-import android.graphics.Rect;
 import android.graphics.drawable.AnimationDrawable;
 import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.ColorDrawable;
 import android.graphics.drawable.Drawable;
-import android.os.Build;
-import android.os.SystemClock;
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
```diff
@@ -85,63 +81,22 @@ static void setPlaceholder(ImageView target, Drawable placeholderDrawable) {
 
     boolean fade = loadedFrom != MEMORY && !noFade;
     if (fade) {
-      this.placeholder = placeholder;
-      animating = true;
-      startTimeMillis = SystemClock.uptimeMillis();
+      startTransition(FADE_DURATION);
+    }
+    else {
```
JakeWharton: This should be on the same line as the `}` above.

dnkoutso:
It used to be like that but decided to use a `BitmapDrawable` to reduce allocations. Perhaps too much micro-optimization especially considered tons of other gains we've had since then.

picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
```diff
@@ -85,63 +81,22 @@ static void setPlaceholder(ImageView target, Drawable placeholderDrawable) {
...
+      // Transition straight to the final drawable
```
JakeWharton: suuuuupernit: period at the end. [Comments should be sentences](http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).

JakeWharton:
LGTM. Just two small nits.

dlew:
Fixed the two nits.

PR #882, opened 2015-02-11
Add volatile to Picasso singleton access.

kevinmmarlow:
- Double-checked locking is broken. See Item 71 in Effective Java.
http://www.oracle.com/technetwork/articles/javase/bloch-effective-08-qa-140880.html


JakeWharton:
Thankfully this probably never matters or will matter due to the nature of how Picasso is used. Good catch though.

PR #887, opened 2015-02-13
Trim url/file path passed in before do anything else

UglyTroLL:
Which should be a right thing to do here in the library. 
The current version cannot handle this properly, e.g. load(" http://square.github.io/picasso/static/sample.png") would fail even it's a valid url.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -295,7 +295,7 @@ public class Picasso {
     if (path.trim().length() == 0) {
       throw new IllegalArgumentException("Path must not be empty.");
     }
-    return load(Uri.parse(path));
+    return load(Uri.parse(path.trim()));
   }
 
   /**
@@ -356,7 +356,10 @@ public class Picasso {
     if (path == null) {
       throw new IllegalArgumentException("path == null");
     }
-    invalidate(Uri.parse(path));
+    if (path.trim().length() == 0) {
+      throw new IllegalArgumentException("Path must not be empty.");
+    }
+    invalidate(Uri.parse(path.trim()));
   }
 
   /**


JakeWharton:
Why is this Picasso's job? Why are you passing in URLs with spaces (which aren't valid URLs)?

UglyTroLL:
Developer could make mistakes, instead of letting them handle this in different places we could just do it here from Picasso itself.

UglyTroLL:
Guess I'm thinking from a different perspective, I've had this problem before when using Picasso.

JakeWharton:
Well then why did your URLs have spaces?

UglyTroLL:
Someone manually copied some url(which had a space before the whole url) from broswer and put it into the databases. Guess my point is, this kind of common mistake happens, and Picasso is the right place to take care of it and make lives easier, even though, technically, it's not Picasso's responsibility to handle it.

crispi7:
Picasso is an image library, if you have spaces it is your problemas, Not Picasso. Anyway, if you want to implement trim un just one place, you should do it on your db method insert.

UglyTroLL:
Yes I totally agree. But this change only makes Picasso a bit safer (or more reliable) to use, and has no negative impact (at least I cannot think of any) on Picasso itself as an image library.
Library users definitely should be responsible for trimming url/file path strings but it doesn't mean Picasso should not do it. I'm just proposing a small and non-harmful change that could make their lives easier.

crispi7:
Yes, Picasso can have it, bit should not. If you are a dev, you Must make high quality code and if you are saving urls that are not valid... Thats not good code.
Having that type of things just make people think the library will make everything for them, and it is not. The library help devs make difficult thing, but dev must create their own good code. Also, maybe is not a heavy task, but if your app needs a los of pictures, it will cost. Furthermore, if you trim urls, a lot of urls will be the same for Picasso, making strange cache/download behaviours

JakeWharton:
I don't think this is an acceptable change. There _is_ a place for being forgiving, but this is not one of them.

The passed in uri is pretty sacred–we use it for quite a lot. Obviously fetching the remote resource, merging requests, cache key, invalidating images from the memory cache, for example. This by itself isn't enough to reject this PR.

The other problem, is actually in something that we're not doing which is automatic URL-encoding. Leading whitespace is usually in error (who would put a space in their scheme?), but trailing whitespace cannot be arbitrarily removed. A uri with a trailing whitespace is completely different than one without. Ideally this would have already been URL-encoded as `%20`, but this is something that can be done by the HTTP client.

PR #876, opened 2015-02-07
Use Gradle for build system

dnkoutso:
Not sure what Mr. Travis wants to work with this.

Also not sure if we want this before 2.5.1 (which might be a small release for 2.5). Doesn't bother me actually.

Tests work fine.

PR #896, opened 2015-02-16
Quit in orderly manner

dnkoutso:
This depends on #876 before it passes.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Dispatcher.java
+++ b/picasso/src/main/java/com/squareup/picasso/Dispatcher.java
@@ -16,12 +16,14 @@
 package com.squareup.picasso;
 
 import android.Manifest;
+import android.annotation.TargetApi;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
+import android.os.Build;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Looper;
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso/Stats.java
```diff
@@ -75,8 +77,13 @@ void dispatchCacheMiss() {
     handler.sendEmptyMessage(CACHE_MISS);
   }
 
+  @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
```
JakeWharton: This isn't how this annotation is meant to be used. You should call a method inside the conditional for the API-specific behavior and then have this annotation on it. With how you have it now you could use API 18+ methods in the `else` block and lint wouldn't complain.

picasso/src/main/java/com/squareup/picasso/Dispatcher.java
```diff
@@ -102,29 +104,34 @@
...
     this.stats = stats;
-    this.batch = new ArrayList<BitmapHunter>(4);
+    this.batch = new ArrayList<>(4);
     this.airplaneMode = Utils.isAirplaneModeOn(this.context);
     this.scansNetworkChanges = hasPermission(context, Manifest.permission.ACCESS_NETWORK_STATE);
     this.receiver = new NetworkBroadcastReceiver(this);
     receiver.register();
   }
 
+  @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
```
JakeWharton: Ditto. Make a `Utils` method that can be shared?

dnkoutso:
Will re-open when we migrate to Gradle

PR #901, opened 2015-02-21
Added ResultCallback interface to return the result Bitmap and LoadedFrom in the callback

carlonzo:
This callback would help to retrieve the Bitmap the user requested after an action fetch(Callback) or into(ImageView, Callback).

Sometime the user can have the necessity to use the requested Bitmap. I know he can achieve this implementing the Target callback, but he would need to create a strong reference to it, and as Jake replied to a guy somewhere here, the Target is intended to be used as an interface to be implemented by a custom view, not as a callback.

Another way to get the Bitmap would be call the method get() after the onSuccess() callback, but it looks a like workaround.

Thanks :)

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Callback.java
+++ b/picasso/src/main/java/com/squareup/picasso/Callback.java
@@ -15,6 +15,8 @@
  */
 package com.squareup.picasso;
 
+import android.graphics.Bitmap;
+
 public interface Callback {
   void onSuccess();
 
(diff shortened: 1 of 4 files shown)


dnkoutso:
This can be done better using our "yet-to-be-developed" MetaData infrastructure for callbacks and transformations. 

Having instanceof checks and casting is a bit ugly if you ask me and serves only this purpose. 

For now closing.

mcuelenaere:
What's the status of this metadata infrastructure?

Now we have to implement this "hackish" solution to get to the bitmap (we also can't implement a custom Target, because only a weak reference is being kept by Picasso and we don't keep a reference ourselves):

```java
final Picasso picasso = Picasso.with(context);
picasso
    .load(bigImageUrl)
    .fetch(new Callback() {
        @Override
        public void onSuccess() {
            try {
                // load again, from cache this time
                Bitmap bitmap = picasso.load(bigImageUrl).get();
                // ...
            } catch (IOException e) {
                // shouldn't happen, unless disk/memory is corrupt
            }
        }

        @Override
        public void onError() {
            // ...
        }
    });
```

carlonzo:
I still think that a solution like the one I suggested can solve in the meantime this missing feature.

JakeWharton:
> What's the status of this metadata infrastructure?

Nonexistent and no one is working on it at this time.

PR #902, opened 2015-02-23
Do not support recycling for remoteviews requests

### LINKED ISSUE #610: Remote views loading doesn't work for RemoteViewsService
Picasso requires all load calls to be made on the main thread, but when using RemoteViewsService (for adapter-like behavior in an app widget) it seems that the RemoteViewsFactory is bound on a non-main thread.  You'd normally create your RemoteViews in RemoteViewFactory.getViewAt() but since that's running on a non-main thread Picasso always fails.

dnkoutso:
#799  removed the `checkMain()` check when submitting RemoteViews requests outside the main thread.

If we want those requests to be submitted outside the main thread then we cant support recycling and must submit straight to the dispatcher instead.

closes #610

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/RequestCreator.java
+++ b/picasso/src/main/java/com/squareup/picasso/RequestCreator.java
@@ -708,6 +708,6 @@ public class RequestCreator {
       action.setImageResource(placeholderResId);
     }
 
-    picasso.enqueueAndSubmit(action);
+    picasso.submit(action);
   }
 }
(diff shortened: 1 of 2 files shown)


JakeWharton:
What effect does this have on list widgets?

dnkoutso:
It probably breaks it entirely. However, the change was already kind of "broken" by allowing bg threads to modifying the targetToAction map.

dnkoutso:
Need a better solution for this entirely.

dnkoutso:
Talked about it, wrong approach. Closing.

Issue #910, opened 2015-02-28
Allow null transformations?


jacobtabak:
Would you merge a PR that allowed you to set a null transformation?  

Currently, you'll encounter the following exception:

>      java.lang.IllegalArgumentException: Transformation must not be null.

Our loading code originally looked something like this:

```
    picasso
        .load(uri)
        .into(target);
```

Now, we're allowing a subclasses to specify an optional transformation, like this:

```
    RequestCreator requestCreator = picasso.load(uri);
    Transformation transformation = getTransformation();
    if (transformation != null) {
      requestCreator.setTransformation(transformation);
    }
    requestCreator.into(target);
```

It would be much nicer if we could instead do this:

```
    picasso
        .load(uri)
        .setTransformation(getTransformation() /* can be null */)
        .into(target);
```

jacobtabak:
Nevermind, I didn't realize you could apply multiple transformations.

JakeWharton:
What about taking a Collection that could be empty?

jacobtabak:
Yes, perhaps a `setTransforms()` or `setTransformations()` would be a bit nicer.  Anything to avoid breaking the builder syntax.

JakeWharton:
We've had this problem as well. Want to PR?

jacobtabak:
Sure, on it

ruckus:
I, too, didnt want to break the builder syntax, so that means always needing to call `Picasso.transform` with a non-null argument. But I always wanted to conditionally apply the transformation.

I solved this by first creating a no-op transformation:

```java
public class NoOpTransformation implements Transformation {
    @Override
    public Bitmap transform(Bitmap source) {
        return source;
    }

    @Override
    public String key() {
        return "noop";
    }
}
```

And then I can keep a local instance variable representing this transformation, otherwise I can re-set the instance variable to my "real" transformation as needed and invocation always look like:

```java
/* somewhere else ... */
backgroundTransformation = new NoOpTransformation();

/* or elsewhere */
backgroundTransformation = new StrokeTransformation();

// later
            mPicasso
                    .load(mMetric.photoURL)
                    .noFade()
                    .transform(backgroundTransformation)
                    .placeholder(R.drawable.grid_gradient)
                    .fit()
                    .into(mViewHolder.background);
```

Are there any downsides to this?

JakeWharton:
Yes. An image loaded with that transformation and an image loaded without
it will have different memory cache keys. I would highly recommend changing
[...]

PR #912, opened 2015-02-28
Overload `transform()` to accept a List parameter

jacobtabak:
Addresses #910 

Unsure about naming of the method here.  

I wanted to do `transform(List)` but then null would match both signatures.

`transforms(List)` wouldn't be grammatically correct.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso/Request.java
@@ -418,6 +418,8 @@ public final class Request {
     /**
      * Add a custom transformation to be applied to the image.
      * <p>
+     * Use {@link #setTransformations(java.util.List)} to set a list of transformations.
+     * <p>
      * Custom transformations will always be run after the built-in transformations.
      */
     public Builder transform(Transformation transformation) {
@@ -433,6 +435,29 @@ public final class Request {
       transformations.add(transformation);
       return this;
     }
+
+
+    /**
+     * Set a list of transformations to be applied to the image.  These transformations will
+     * overwrite any existing transformations.
+     * <p>
+     * Use {@link #transform(Transformation)} to add a transformation without overwriting existing
+     * transformations.
+     * <p>
+     * Custom transformations will always be run after the built-in transformations.
+     */
+    public Builder setTransformations(List<Transformation> transformations) {
+      if (transformations != null) {
+        this.transformations = new ArrayList<Transformation>(transformations.size());
+        for (Transformation transformation : transformations) {
+          transform(transformation);
+        }
+      } else {
+        this.transformations = null;
+      }
+      return this;
+    }
+
 
     /** Create the immutable {@link Request} object. */
     public Request build() {
--- a/picasso/src/main/java/com/squareup/picasso/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso/Request.java
@@ -418,8 +418,6 @@ public final class Request {
     /**
      * Add a custom transformation to be applied to the image.
      * <p>
-     * Use {@link #setTransformations(java.util.List)} to set a list of transformations.
-     * <p>
      * Custom transformations will always be run after the built-in transformations.
      */
     public Builder transform(Transformation transformation) {
@@ -436,27 +434,20 @@ public final class Request {
       return this;
     }
 
-
     /**
-     * Set a list of transformations to be applied to the image.  These transformations will
-     * overwrite any existing transformations.
-     * <p>
-     * Use {@link #transform(Transformation)} to add a transformation without overwriting existing
-     * transformations.
+     * Add a list of custom transformations to be applied to the image.
      * <p>
      * Custom transformations will always be run after the built-in transformations.
      */
-    public Builder setTransformations(List<Transformation> transformations) {
+    public Builder transform(List<? extends Transformation> transformations) {
       if (transformations == null) {
         throw new IllegalArgumentException("Transformation list must not be null.");
       }
-      this.transformations = new ArrayList<Transformation>(transformations.size());
-      for (Transformation transformation : transformations) {
-        transform(transformation);
+      for (Object transformation : transformations) {
+        transform((Transformation) transformation);
       }
       return this;
     }
-
 
     /** Create the immutable {@link Request} object. */
     public Request build() {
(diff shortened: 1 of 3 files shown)


picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -434,6 +436,29 @@ public Builder transform(Transformation transformation) {
...
+     * Set a list of transformations to be applied to the image.  These transformations will
+     * overwrite any existing transformations.
+     * <p>
+     * Use {@link #transform(Transformation)} to add a transformation without overwriting existing
+     * transformations.
+     * <p>
+     * Custom transformations will always be run after the built-in transformations.
+     */
+    public Builder setTransformations(List<Transformation> transformations) {
+      if (transformations != null) {
```
JakeWharton: Null is an error. Throw.

picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -434,6 +436,29 @@ public Builder transform(Transformation transformation) {
...
+    /**
+     * Set a list of transformations to be applied to the image.  These transformations will
+     * overwrite any existing transformations.
+     * <p>
+     * Use {@link #transform(Transformation)} to add a transformation without overwriting existing
+     * transformations.
+     * <p>
+     * Custom transformations will always be run after the built-in transformations.
+     */
+    public Builder setTransformations(List<Transformation> transformations) {
```
JakeWharton: Generic should be `? Super Transformation` so someone could pass `List<MyTransformation>`

picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -434,6 +436,29 @@ public Builder transform(Transformation transformation) {
       return this;
     }
 
+
+    /**
+     * Set a list of transformations to be applied to the image.  These transformations will
+     * overwrite any existing transformations.
```
JakeWharton: Is this the behavior we want? Or should it just append? I would have thought the latter (which probably would require it to be named `addTransformations`)

jacobtabak: I thought it was a bit odd that you can dump these transformations into the `RequestCreator` but you can't remove them or access them afterwards.  This approach allows you to take an existing request creator and remove any transformations.

JakeWharton: Do you have an example of when that would be useful?

jacobtabak: Not really, I'm fine with doing it how you suggested.

jacobtabak:
Since neither the single Transformation or the list of Transformations can be null, we can overload the `.transform()` signature to accept either a single Transformation or a list.

jacobtabak:
Ok, changed things up a bit. I removed the test for the IllegalArgumentException on transform(null) since it won't even compile anymore.

JakeWharton:
Duplicate the test and add explicit casts from null for each overload.

jacobtabak:
Updated the tests.

picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -434,6 +434,21 @@ public Builder transform(Transformation transformation) {
       return this;
     }
 
+    /**
+     * Add a list of custom transformations to be applied to the image.
+     * <p>
+     * Custom transformations will always be run after the built-in transformations.
+     */
+    public Builder transform(List<? extends Transformation> transformations) {
```
JakeWharton: You want `List<? super Transformation>` here. Then you get `Transformation` objects without casting.

picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
```diff
@@ -790,6 +797,28 @@ public void nullKeyTransformationInvalid() {
...
+        return null;
+      }
+    });
+    new RequestCreator().transform(transformations);
+  }
+
+  @Test
+  public void transformationListImplementationValid() {
+    List<TestTransformation> transformations = new ArrayList<TestTransformation>();
+    transformations.add(new TestTransformation("test"));
```
JakeWharton: Use `Collections.singletonList` or `Arrays.asList` to easily create this guy.

picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
```diff
@@ -790,6 +797,28 @@ public void nullKeyTransformationInvalid() {
...
+    new RequestCreator().transform(transformations);
```
JakeWharton: This test doesn't verify anything. We need to at least assert that each transformation in the list was included in the final request in the same order.

picasso/src/main/java/com/squareup/picasso/Request.java
jacobtabak: Sorry - I added the cast because I took your advice and tried it with `<? super Transformation>` first.  The test I added didn't compile, so I changed it to `<? extends Transformation>`.  So I just removed the unnecessary cast and it should be good now.

JakeWharton: Yes, I think I have them backwards in my head (again)...

picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
jacobtabak: Ironically this test actually caught a bug when I wrote it originally.  I had forgot the `this.` in `this.transformations` in some code that's no longer part of the PR.  

I put a bit of thought into the quality of the tests.  There are no existing tests that checked if `.transform()` actually added the transformation.  `Request.Builder data` is a private field in `RequestCreator`, and `List<Transformation> transformations` in `Request.Builder` is as well,  and I assumed that you guys would be reluctant to increase visibility.  

But I'd definitely feel better about increasing visibility and writing proper tests.

JakeWharton: Yeah our tests... are a mess. @dnkoutso and I regularly bounce back complaints to each other about it. They way too white-box and tiny in scope.

I'll think about this.

dnkoutso:
Lets hold off on this one for 2.6.

JakeWharton:
I squashed and merged this because I don't think we should block it on our ability to test. Plus I want to use it!

Issue #936, opened 2015-03-19
Transparent images overlaid in ListView adapter


JayNewstrom:
I have a ListView with an adapter which uses picasso to set transparent images, after scrolling (view reuse) the images appear on top of the one previously used for that row. As you can see in the screen shot, I have scrolled up and down on the top half, to show overlapping, the bottom rows haven't been reused because I didn't scroll them off screen.
![screenshot_2015-03-19-15-09-52](https://cloud.githubusercontent.com/assets/713288/6739925/3f505fe4-ce4a-11e4-8f83-07c25a013c7b.png)

JakeWharton:
List version and show adapter code.

JayNewstrom:
Just started happening on 2.5.1 it's very simple adapter code. Just picasso.load(url).into(imageView); I can put together a sample app if need be.

JakeWharton:
Probably due to the switch to `TransitionDrawable` in #855. cc/ @dlew

JakeWharton:
Are you using placeholders?

JayNewstrom:
Not in the place that it is happening.

dlew:
It could definitely be related. Which OS version? Maybe `TransitionDrawable` won't work so well on older platforms.

JayNewstrom:
I've only tried it on 4.4, 5.0, 5.1 and it's happening on all of them.

dlew:
Could you give a sample of it happening?

JayNewstrom:
https://github.com/JayNewstrom/picassoOverlay

dlew:
Thanks! I'll take a look soon.

dlew:
Early results: `TransitionDrawable.setCrossFadeEnabled(true)` in `PicassoDrawable`'s constructor fixes the issue.

Not 100% sure why yet; going to keep digging in until I know for sure what's going on.

dlew:
Aha! I see the second problem, which is larger than the first. It's that `PicassoDrawable` uses the `ImageView`'s current drawable as a placeholder. Since the `ImageView` is being recycled that means it's using a `PicassoDrawable` as the placeholder for another `PicassoDrawable` and layered drawable madness ensues.

We just need to prevent `PicassoDrawable` from using itself as a placeholder. It should not only fix this problem but there are also performance implications it causes for opaque images (since it'll be drawing previous layers behind it).

We still need to enable cross fades, otherwise both the placeholder *and* the bitmap render after the transition is completely (which would cause transparent images to fail).

[...]

PR #944, opened 2015-03-20
Fixed issues w/ PicassoDrawable + transparent images

dlew:
Two problems were occurring at once which would cause transparent
images to render incorrectly.

1. Recycling of ImageViews. If no placeholder was set, then Picasso
   reuses the previous drawable of the ImageView as a placeholder.
   This causes problems when the previous placeholder is a
   LayerDrawable, since the layers will just pile up.

   Theoretically this behavior might be correct, but I doubt anyone
   is interested in the performance nightmare it could present
   with lots of recycling and no explicit placeholders. Instead,
   we only use the placeholder layer of previous PicassoDrawables.

2. Lack of cross-fade. If a TransitionDrawable is not set to cross
   fade, then after it's done animating it always draws *both*
   layers, where we actually only want the 2nd layer drawn.

Addresses #936

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
@@ -44,6 +44,12 @@ final class PicassoDrawable extends TransitionDrawable {
   static void setBitmap(ImageView target, Context context, Bitmap bitmap,
       Picasso.LoadedFrom loadedFrom, boolean noFade, boolean debugging) {
     Drawable placeholder = target.getDrawable();
+
+    // Avoid infinite nesting of placeholders.
+    if (placeholder instanceof PicassoDrawable) {
+      placeholder = ((PicassoDrawable) placeholder).getDrawable(0);
+    }
+
     if (placeholder instanceof AnimationDrawable) {
       ((AnimationDrawable) placeholder).stop();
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
```diff
@@ -74,6 +80,8 @@ static void setPlaceholder(ImageView target, Drawable placeholderDrawable) {
       new BitmapDrawable(context.getResources(), bitmap)
     });
 
+    setCrossFadeEnabled(true);
```
dnkoutso: Should this always be set or should we check the `fade` flag below

```java
    if (fade) {
      startTransition(FADE_DURATION);
    } else {
      // Transition straight to the final drawable.
      startTransition(0);
    }
```

dlew: In the case of `TransitionDrawable` it always has to be set; otherwise it'll draw both layers when the "transition" is done.

dlew:
I'm going close this and I'll open a revert PR. As it stands, a non-fading `TransitionDrawable` still has to draw twice, which isn't great (and I didn't realize until I really dug into the code).

I'll have to think through a better solution to #427.

Issue #943, opened 2015-03-20
Picasso with RoundedImageView problem after lastest update


MoshDev:
Am using Picasso with RoundedImageView library, the day before it was working fine, and today seems there is an update on Picasso and now each image loading throws a NullPointerException.

```
    java.lang.NullPointerException
            at android.graphics.drawable.LayerDrawable$LayerState.<init>(LayerDrawable.java:641)
            at android.graphics.drawable.LayerDrawable.mutate(LayerDrawable.java:578)
            at com.makeramen.roundedimageview.RoundedImageView.applyColorMod(RoundedImageView.java:285)
            at com.makeramen.roundedimageview.RoundedImageView.updateAttrs(RoundedImageView.java:307)
            at com.makeramen.roundedimageview.RoundedImageView.updateAttrs(RoundedImageView.java:312)
            at com.makeramen.roundedimageview.RoundedImageView.updateDrawableAttrs(RoundedImageView.java:258)
            at com.makeramen.roundedimageview.RoundedImageView.setImageDrawable(RoundedImageView.java:207)
            at com.squareup.picasso.PicassoDrawable.setBitmap(PicassoDrawable.java:52)
            at com.squareup.picasso.ImageViewAction.complete(ImageViewAction.java:48)
            at com.squareup.picasso.Picasso.deliverAction(Picasso.java:558)
            at com.squareup.picasso.Picasso.complete(Picasso.java:510)
            at com.squareup.picasso.PicassohandleMessage(Picasso.java:117)
            at android.os.Handler.dispatchMessage(Handler.java:99)
            at android.os.Looper.loop(Looper.java:137)
            at android.app.ActivityThread.main(ActivityThread.java:5041)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:511)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:793)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:560)
            at dalvik.system.NativeStart.main(Native Method)
```

dnkoutso:
Most likely related to https://github.com/square/picasso/pull/855

ChrisMCMine:
Same problem with Picasso 2.5.1 and the [Shape ImageView library](https://github.com/siyamed/android-shape-imageview)

dnkoutso:
Please fallback to 2.5.0 for now we are investigating on the issues.

[...]

Issue #939, opened 2015-03-20
Picasso 2.5.1 java.lang.StackOverflowError


AlexYXB:
```
03-20 12:29:04.830: E/ACRA(23613): ACRA caught a StackOverflowError exception for wwface.android.teacher. Building report.
03-20 12:29:05.955: E/UI(23613): CrashReport:::java.lang.StackOverflowError
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
03-20 12:29:05.955: E/UI(23613): 	at android.graphics.drawable.TransitionDrawable.draw(TransitionDrawable.java:208)
03-20 12:29:05.955: E/UI(23613): 	at com.squareup.picasso.PicassoDrawable.draw(PicassoDrawable.java:92)
[...]

PR #945, opened 2015-03-20
Revert "Use TransitionDrawable for PicassoDrawable"

### LINKED ISSUE #936: Transparent images overlaid in ListView adapter

dlew:
This reverts commit 02cd6de9ab1a88e018096c79de52550238084bef.

This will fix #936, #939, and #943.


JakeWharton:
¯\\\_(ツ)\_/¯

PR #967, opened 2015-04-08
Added support for notification tags.

dzeikei:
From http://developer.android.com/reference/android/app/NotificationManager.html
>Each of the notify methods takes an int id parameter and optionally a String tag parameter, which may be null. These parameters are used to form a pair (tag, id), or (null, id) if tag is unspecified. This pair identifies this notification from your app to the system, so that pair should be unique within your app.

Current Picasso API supports referring to a notification only by the notification id which means you can't find the correct notification if you're using the optional notification tags.

There is also an overloaded method without the optional tag for API backwards-compatibility.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/RemoteViewsAction.java
+++ b/picasso/src/main/java/com/squareup/picasso/RemoteViewsAction.java
@@ -102,19 +102,21 @@ abstract class RemoteViewsAction extends Action<RemoteViewsAction.RemoteViewsTar
 
   static class NotificationAction extends RemoteViewsAction {
     private final int notificationId;
+    private final String notificationTag;
     private final Notification notification;
 
     NotificationAction(Picasso picasso, Request data, RemoteViews remoteViews, int viewId,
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso/RemoteViewsAction.java
```diff
@@ -102,19 +102,21 @@ void setImageResource(int resId) {
 
   static class NotificationAction extends RemoteViewsAction {
     private final int notificationId;
+    private final String notificationTag;
```
dnkoutso: I think you need to update equals/hashCode to include this field.

dnkoutso: Or maybe not...

dnkoutso: Nope I think you are OK without it.

dnkoutso:
Have you signed the CLA?

dzeikei:
I have now :)

dnkoutso:
Thanks!

PR #977, opened 2015-04-13
Expose BitmapFactory.Options.inPurgeable through RequestCreator.purgeable()

sneuberger-amazon:
When purgeable() is called, we set inPurgeable and inInputShareable to true. We
also decode the bitmap from a byte array instead of a stream, because
inPurgeable only affects the fomer (we only do this for API level <21 because
inPurgeable has no effect for API level 21+).

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -120,10 +120,12 @@ class BitmapHunter implements Runnable {
     final boolean calculateSize = RequestHandler.requiresInSampleSize(options);
 
     boolean isWebPFile = Utils.isWebPFile(stream);
+    boolean isPurgeable = request.purgeable && android.os.Build.VERSION.SDK_INT < 21;
     markStream.reset(mark);
-    // When decode WebP network stream, BitmapFactory throw JNI Exception and make app crash.
-    // Decode byte array instead
-    if (isWebPFile) {
+    // We decode from a byte array because, a) when decoding a WebP network stream, BitmapFactory
+    // throws a JNI Exception, so we workaround by decoding a byte array, or b) user requested
+    // purgeable, which only affects bitmaps decoded from byte arrays.
+    if (isWebPFile || isPurgeable) {
       byte[] bytes = Utils.toByteArray(stream);
       if (calculateSize) {
         BitmapFactory.decodeByteArray(bytes, 0, bytes.length, options);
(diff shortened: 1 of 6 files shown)


picasso/src/main/java/com/squareup/picasso/RequestCreator.java
```diff
@@ -366,6 +366,16 @@ public RequestCreator networkPolicy(NetworkPolicy policy, NetworkPolicy... addit
     return this;
   }
 
+  /** Set inPurgeable and inInputShareable when decoding. This will force the bitmap to be decoded
```
dnkoutso: nit: need extra line here.

dnkoutso:
checkstyle fails in the build, also need a test for this.

sneuberger-amazon:
1. Fixed a bug where bitmap options could be null (I originally wrote this patch for an old version of Picasso where bitmap options was never null).
2. Wrote tests. I don't have a test for BitmapHunter.decodeStream(). The only way I see to test it would be to extract two methods decodeFromArray() and decodeFromStream() and using a spy() to verify which one is called. Let me know if you want this.
3. Fixed style issues. Build now passes.

sneuberger-amazon:
Is there any more work I can do on this?

sneuberger-amazon:
Friendly ping, @dnkoutso or any other maintainer who can take a look at this.

picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
```diff
@@ -120,10 +120,12 @@ static Bitmap decodeStream(InputStream stream, Request request) throws IOExcepti
     final boolean calculateSize = RequestHandler.requiresInSampleSize(options);
 
     boolean isWebPFile = Utils.isWebPFile(stream);
+    boolean isPurgeable = request.purgeable && android.os.Build.VERSION.SDK_INT < 21;
```
dnkoutso: sucks we still cant use the constant since we are not on gradle.

picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -79,6 +79,7 @@
   public final float rotationPivotY;
   /** Whether or not {@link #rotationPivotX} and {@link #rotationPivotY} are set. */
   public final boolean hasRotationPivot;
+  public final boolean purgeable;
```
dnkoutso: nit: add a comment here.

dnkoutso:
LGTM. Want to address the minor nits and I will merge, have you signed CLA?

sneuberger-amazon:
Added the comment, and a couple minor fixes. I have signed the CLA.

ypresto:
Note that this feature should not be used in kitkat because there is nasty file descriptor leak bug in BitmapFactory...

https://code.google.com/p/android/issues/detail?id=65638#makechanges

PR #982, opened 2015-04-20
Change picasso to accept multiple RequestTransformers registartion

daisy1754:
Keep existing method Picasso.Builder#requestTransformer for backword compatibility, add Picasso.Builder#addRequestTransformer for registering potentially multiple RequestTransformer.

User of library may fetch images from multiple sources and hence have multiple rewrite rules to apply. This PR allows library users to have multiple RequestTransformers with a single responsibility for each.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -85,13 +85,6 @@ public class Picasso {
      * @return The original request or a new request to replace it. Must not be null.
      */
     Request transformRequest(Request request);
-
-    /** A {@link RequestTransformer} which returns the original request. */
-    RequestTransformer IDENTITY = new RequestTransformer() {
-      @Override public Request transformRequest(Request request) {
-        return request;
-      }
-    };
   }
 
   /**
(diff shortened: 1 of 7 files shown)


JakeWharton:
Would you actually use multiple? It's also worth noting that you could implement a single instance that simply runs multiple itself as an existing way of accomplishing this.

daisy1754:
Yes, we have a single top-level RequestTransformer class that delegates to multiple RequestTransformers (for handling several urls), and I was thinking library should allow multiple RequestTransformers as it allow multiple RequestHandlers.

JakeWharton:
Can you explain what each individual one does? I'm just trying to understand exactly what you are doing.

daisy1754:
One is for transforming url for CDNs, and another is for transforming url for local file system.

JakeWharton:
I'm sorry for the ridiculous delay here. We added this feature and it will be available in Picasso 3.0, if you're still even a Picasso user and choose to upgrade.

daisy1754:
Thanks for response anyway. Glad to see Picasso is keep evolving. I don't involved in Android app development anymore but if I need to pick image library for Android in the future Picasso will be my first choice :)

PR #1015, opened 2015-05-08
Disallow setting singleton instance to null

f2prateek:


f2prateek:
Also reformatted two lines that were exceeding the 100 chars.

PR #1021, opened 2015-05-14
EXIFOrientation Handling improvements

JohnWowUs:
closes #1020

PR #692, opened 2014-10-07
Support EXIF orientation for network JPG/TIFF

jacobtabak:
Uses commons-imaging to determine orientation (only if it's on the classpath, it's an optional dependency).  Tests intentionally omitted for now, happy to add them if you want this feature.

Since commons-imaging is still at 1.0.0-SNAPSHOT I had to temporarily add the Apache snapshot repository but this API hasn't changed since it was brought into commons.

PR #1031, opened 2015-05-19
Lightweight EXIF orientation for network JPG/TIFF

JohnWowUs:
I took some of the comments in PR #692 into account and came with a new solution light weight (no dependencies) for EXIF support for network JPG/TIFF files. These integrate the changes made in PR #1021 to handle all EXIF orientations.

I had to add some new resources for the tests hopefully that won't cause any problems with Travis.


picasso/src/main/java/com/squareup/picasso/ExifStreamReader.java
```diff
@@ -0,0 +1,158 @@
...
+ * Adapted for picasso 
+ */
+
+package com.squareup.picasso;
+
+import android.util.Log;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class ExifStreamReader {
```
JakeWharton: not public, +final

picasso/src/main/java/com/squareup/picasso/ExifStreamReader.java
```diff
@@ -0,0 +1,158 @@
...
+  private static final String TAG = "CameraExif";
+
+  // Returns the orientation value
+  public static int getOrientation(InputStream stream) throws IOException {
+    if (stream == null) {
+      return 0 ;
+    }
+    MarkableInputStream markStream = new MarkableInputStream(stream);
+    long mark = markStream.savePosition(65536); 
+    byte[] header = new byte[65536];
```
JakeWharton: We can't do this kind of allocation. I have a better solution in mind but not the time to implement it. I have a flight on Sunday. I'll try to do it then.

picasso/src/main/java/com/squareup/picasso/NetworkRequestHandler.java
```diff
@@ -26,6 +26,7 @@
 
 class NetworkRequestHandler extends RequestHandler {
   static final int RETRY_COUNT = 2;
+  private static final int MARKER = 65536;
```
JakeWharton: Unused

picasso/src/main/java/com/squareup/picasso/NetworkRequestHandler.java
```diff
@@ -69,6 +70,12 @@ public NetworkRequestHandler(Downloader downloader, Stats stats) {
     if (loadedFrom == NETWORK && response.getContentLength() > 0) {
       stats.dispatchDownloadFinished(response.getContentLength());
     }
+    try {
+      int orientation = ExifStreamReader.getOrientation(is);
+      return new Result(is, loadedFrom, orientation);
+    } catch (IOException ignored) {
+
```
JakeWharton: Remove empty line.

JakeWharton:
Lovely test cases!

picasso/src/main/java/com/squareup/picasso/ExifStreamReader.java
JakeWharton: For what it's worth, we can merge this as-is once all the other things are cleaned up and I'll fix this after.

JohnWowUs:
The tests works locally but not with Travis for some reason.

JakeWharton:
I can try locally later tonight.

picasso/src/test/java/com/squareup/picasso/ExifStreamReaderTest.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (C) 2013 Square, Inc.
```
dnkoutso: 2015

JakeWharton:
The test fails locally for me. I'll add some logging to the parsing and see what's wrong at lunch.

JohnWowUs:
The res folder in the test folder should be resources.

dnkoutso:
squash to one commit.

dnkoutso:
Holding on on this one for perhaps the next release. 2.5.3 is supposed to be a small release and its turning out to be pretty big.

dkharrat:
@JohnWowUs I just tried your branch to test a similar issue I found, but it didn't work in that images are broken and don't load at all. I saw the following messages in the logs:

    D/skia﹕ --- SkImageDecoder::Factory returned null

After investigating it, it seems like it's due to the InputStream already being consumed by `ExifStreamReader.getOrientation`.

picasso/src/main/java/com/squareup/picasso/ExifStreamReader.java
```diff
@@ -0,0 +1,165 @@
...
+
+final class ExifStreamReader {
+  private static final String TAG = "CameraExif";
+
+  // Returns the orientation value
+  public static int getOrientation(InputStream stream) throws IOException {
+    if (stream == null) {
+      return 0;
+    }
+    MarkableInputStream markStream = new MarkableInputStream(stream);
```
dkharrat: Isn't using a `MarkableInputStream` here irrelevant, since it's local to `ExifStreamReader` only and will be discarded at the end of the call. The `InputStream` provided by the caller will still have been consumed and is independent of `MarkableInputStream`. Instead, I think you want to find a way to maintain the original `InputStream` (perhaps best done by the caller?)

I believe this is the root cause of the issue I was having using your branch, where images don't load up.

JohnWowUs: The MarkableInputStream is used to allow us to reset the Inputstream i.e. not have it consumed.

icesmith:
I tried to build JohnWowUs:IFStream-EXIF branch, and have the save issue. It would be great, if you can fix it and merge into picasso.

picasso/src/main/java/com/squareup/picasso/ExifStreamReader.java
icesmith: I tried to build JohnWowUs:IFStream-EXIF branch, and have the save issue. It would be great, if you can fix it and merge into picasso.

dkharrat: @JohnWowUs I see what you mean; I was under the impression that the `MarkableInputStream` maintains its own marker and resetting it just resets `MarkableInputStream`, but after looking at the code, I see that it also tries to reset the original `InputStream`.

However, it seems that it does not work (hence, why I'm seeing failures in decoding the images). I believe it happens when the downloader is `UrlConnectionDownloader`, since that downloader uses the `InputStream` provided by `openConnection()`, which doesn't seem resettable.

Can you also try it on your end to confirm?

icesmith: I use OkHttpDownloader. Also I use NetworkInterceptor, wich replaces original InputStream with new one, because original stream is encrypted and I need to decrypt it.

JohnWowUs:
@icesmith @dkharrat 
I believe it should be fixed now. Could you test on your end?

icesmith:
Great!, It works for me now, thank you @JohnWowUs

dkharrat:
@JohnWowUs it also works for me now. Thanks for fixing it!

Shusshu:
Is this going to be included in 2.5.3 ?

abhinavraj:
Any update on timeline for merging this?

bychkovdmitrii:
Any chance this is going to be merged soon?

Issue #965, opened 2015-04-04
Picasso doesn't load images and never invoke callback


fraggjkee:
Hello guys, I have a `ListView` and the following code in its adapter:

```
...

LayoutInflater layoutInflater = LayoutInflater.from(mContext);
final ViewHolder viewHolder;
if (convertView == null) {
      convertView = layoutInflater.inflate(R.layout.card_view_item_order, null);
      viewHolder = new ViewHolder();
      viewHolder.mImageView = (ImageView) convertView.findViewById(R.id.card_view_item_order_image_view);
      convertView.setTag(viewHolder);
} else {
      viewHolder = (ViewHolder) convertView.getTag();
}
Picasso.with(mContext).load(url).into(viewHolder.mImageView);

...

private static class ViewHolder {
      ImageView mImageView;
}
```

For some reason Picasso doesn't load images and also never invokes a callback if I pass it along with `ImageView` (strong reference to callback is in place), neither `onSuccess` nor `onError`. I was using Picasso in several projects before and it used to work perfectly, but for some reason it doesn't work anymore... The URL is correct and other libraries and low-level code including `InputStream` and related stuff are working well for this specific case. Any ideas? If it helps, `ImageView` has a fixed height = 100dp and MATCH_PARENT width. It is also embedded into `CardView` widget.

**Version**: 2.5.2
**Logs**:
>04-04 11:28:08.403    3736-3736/com.test D/Picasso﹕ Main        created      [R0] Request{http://damp-crag-3442.herokuapp.com/assets/android_order_item_img-b2596baa04c79bc61f403f34528edc60.png}
04-04 11:28:08.409    3736-3736/com.test D/Picasso﹕ Main        created      [R1] Request{http://damp-crag-3442.herokuapp.com/assets/android_order_item_img-b2596baa04c79bc61f403f34528edc60.png}
04-04 11:28:08.409    3736-3755/com.test D/Picasso﹕ Dispatcher  enqueued     [R0]+6ms
04-04 11:28:08.409    3736-3755/com.test D/Picasso﹕ Hunter      joined       [R1]+0ms to [R0]+6ms, [R1]+0ms
04-04 11:28:08.410    3736-3736/com.test D/Picasso﹕ Main        created      [R2] Request{http://damp-crag-3442.herokuapp.com/assets/android_order_item_img-b2596baa04c79bc61f403f34528edc60.png}
[...]

Issue #907, opened 2015-02-27
Picasso 2.5.0 does not display big images


darzul:
Hello,

I have some heavy pictures, about 10Mo for each one. On Picasso 2.4.0 there is no problems to display them, but with Picasso 2.5.0 any pictures are displayed.

Here is how I display the pictures: 

mPicasso.load(file)
               .error(R.drawable.ic_launcher)
               .resize(mScreenWidth, 0)
               .into(iv);

Here are the log for one image:

D/Picasso﹕ Main        created      [R5] Request{file:///storage/emulated/0/Pictures/Weekend en Corse/DSC_0008.JPG resize(1080,0)}
D/Picasso﹕ Dispatcher  enqueued     [R5]+34ms
D/Picasso﹕ Dispatcher  enqueued     [R5]+32ms
D/Picasso﹕ Dispatcher  batched        [R5]+556ms for error
D/Picasso﹕ Main           errored         [R5]+748ms
D/Picasso﹕ Dispatcher  delivered      [R5]+766ms

I have another set of pictures which are more little and there is no problem on Picasso 2.5.0 (3Mo per picture).

Here is a picture impossible to display:

![dsc_0110](https://cloud.githubusercontent.com/assets/4621928/6416784/0ee589a2-bea9-11e4-9604-f49b4071c2cc.JPG)

dnkoutso:
Attach a listener please, `picasso.setListener(new Listener() { exception .printStackTrace() }`

darzul:
Could you be more specific please ?

I don't find setListener() method

dnkoutso:
Apologies, I wrote this rather quickly:

```java
Picasso picasso = new Picasso.Builder(context).listener(new Picasso.Listener() {
      @Override public void onImageLoadFailed(Picasso picasso, Uri uri, Exception exception) {
        exception.printStackTrace();
      }
    }).build();
```

Then use that instance for the loading part instead.

darzul:
W/System.err﹕ java.io.IOException: Cannot reset
W/System.err﹕ at com.squareup.picasso.MarkableInputStream.reset(MarkableInputStream.java:99)
W/System.err﹕ at com.squareup.picasso.BitmapHunter.decodeStream(BitmapHunter.java:140)
W/System.err﹕ at com.squareup.picasso.BitmapHunter.hunt(BitmapHunter.java:217)
W/System.err﹕ at com.squareup.picasso.BitmapHunter.run(BitmapHunter.java:159)
[...]

Issue #881, opened 2015-02-11
Picasso not loading some images when using fit()


yychun1217:
I am using Picasso (Version: 2.5.0) for my project. In my listView, there are images to be loaded from online. I used fit() & centerInside() to resize my images. Most of the images are loaded successfully, while a few get an error. I have checked that the url contains no special characters, and they are of acceptable length.

My code is as below:
Picasso.with(mContext) 
.load(UrlEncoder.encode(interiorDesign.getBrand_image_url())) 
.config(Bitmap.Config.RGB_565)
.error(R.drawable.blank)
.fit() 
.centerInside() 
.into(holder.brand); 

In fact, I encounter the same problem when using fit() & centerCrop().

For your reference, Here is my question in Stackoverflow.
Here: http://stackoverflow.com/questions/28426468/picasso-cannot-load-images-for-some-url-no-special-characters?noredirect=1#comment45189039_28426468

dnkoutso:
Known issue: https://github.com/square/picasso/issues/465 will be fixed in 2.5.1.

andrejvanderzee:
Looks like same issue I am seeing with large images:

02-28 22:17:02.152    6367-6367/com.fansz.prestaging E/FanszApplication﹕ http://staging.assets.fansz.com/uploads/f92/c/234b40c5bd9346818f2f7f0e9726cf92/img_20150227_150634.jpg
    java.io.IOException: Cannot reset
            at com.squareup.picasso.MarkableInputStream.reset(MarkableInputStream.java:99)
            at com.squareup.picasso.BitmapHunter.decodeStream(BitmapHunter.java:140)
            at com.squareup.picasso.BitmapHunter.hunt(BitmapHunter.java:217)
            at com.squareup.picasso.BitmapHunter.run(BitmapHunter.java:159)
            at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
            at java.util.concurrent.FutureTask.run(FutureTask.java:237)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
            at java.lang.Thread.run(Thread.java:818)

dnkoutso:
We are on it. thanks for the report.

andrejvanderzee:
[...]

Issue #686, opened 2014-10-04
Images that have color profile fail to decode when using fit()


muteor:
Hi,

I am having trouble with a few images when using fit() I think this may have something to do with them having icc color profiles as that is the only difference I see from jpgs that load.

The images are:

http://fc06.deviantart.net/fs71/f/2014/267/2/6/catharsis_by_mimikascraftroom-d80cw05.jpg
http://fc08.deviantart.net/fs70/f/2014/241/d/a/porec_dressed_in_blue_by_r_maric-d7x4cdp.jpg
http://fc09.deviantart.net/fs70/f/2014/276/f/e/rock___by_glooh-d81fggf.jpg
http://fc02.deviantart.net/fs70/f/2014/274/f/2/mig_29_by_vipmig-d819sv8.jpg

If I dont use fit on these images they are rendered into the imageview correctly, so android can display them, but maybe the bitmap factory is failing or some other part of picasso. In any case seems that we end up with a null result bitmap after resize.

Tried to include as much info as possible, I have sample project if you need it.

Thanks

Keith

Here is the log:

```
10-04 12:18:42.079    1393-1393/com.muteor.picassobug D/Picasso﹕ Main        created      [R1] Request{http://fc06.deviantart.net/fs71/f/2014/267/2/6/catharsis_by_mimikascraftroom-d80cw05.jpg resize(400,400)}
10-04 12:18:42.079    1393-1407/com.muteor.picassobug D/Picasso﹕ Dispatcher  enqueued     [R1]+0ms
10-04 12:18:42.079    1393-1413/com.muteor.picassobug D/Picasso﹕ Hunter      executing    [R1]+1ms
10-04 12:18:43.479    1393-1407/com.muteor.picassobug D/Picasso﹕ Dispatcher  retrying     [R1]+1400ms
10-04 12:18:43.479    1393-1415/com.muteor.picassobug D/Picasso﹕ Hunter      executing    [R1]+1400ms
10-04 12:18:43.487    1393-1415/com.muteor.picassobug D/dalvikvm﹕ GC_FOR_ALLOC freed 3542K, 51% free 3599K/7264K, paused 4ms, total 4ms
10-04 12:18:43.991    1393-1407/com.muteor.picassobug D/Picasso﹕ Dispatcher  retrying     [R1]+1911ms
10-04 12:18:43.991    1393-1416/com.muteor.picassobug D/Picasso﹕ Hunter      executing    [R1]+1912ms
[...]

PR #1035, opened 2015-05-21
Added dynamic limit option to MarkableInputStream

JohnWowUs:
A fix for the problem encountered in issues #364, #465, #626, #686, #881, #907 and #965. The approach I took was to automatically increase the read limit if a read from the stream would exceed the current limit.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -121,8 +121,8 @@ class BitmapHunter implements Runnable {
   static Bitmap decodeStream(InputStream stream, Request request) throws IOException {
     MarkableInputStream markStream = new MarkableInputStream(stream);
     stream = markStream;
-
-    long mark = markStream.savePosition(65536); // TODO fix this crap.
+    markStream.allowMarksToExpire(false);
+    long mark = markStream.savePosition(1024);
 
     final BitmapFactory.Options options = RequestHandler.createBitmapOptions(request);
     final boolean calculateSize = RequestHandler.requiresInSampleSize(options);
@@ -146,9 +146,9 @@ class BitmapHunter implements Runnable {
         BitmapFactory.decodeStream(stream, null, options);
         RequestHandler.calculateInSampleSize(request.targetWidth, request.targetHeight, options,
             request);
-
         markStream.reset(mark);
       }
+      markStream.allowMarksToExpire(true);
       Bitmap bitmap = BitmapFactory.decodeStream(stream, null, options);
       if (bitmap == null) {
         // Treat null as an IO exception, we will eventually retry.
--- a/picasso/src/main/java/com/squareup/picasso/MarkableInputStream.java
+++ b/picasso/src/main/java/com/squareup/picasso/MarkableInputStream.java
@@ -26,23 +26,30 @@ import java.io.InputStream;
  */
 final class MarkableInputStream extends InputStream {
   private static final int DEFAULT_BUFFER_SIZE = 4096;
-
+  private static final int DEFAULT_LIMIT_INCREMENT = 1024;
   private final InputStream in;
 
   private long offset;
   private long reset;
   private long limit;
   private long defaultMark = -1;
+  private boolean allowExpire = true;
+  private int limitIncrement = -1;
 
   public MarkableInputStream(InputStream in) {
     this(in, DEFAULT_BUFFER_SIZE);
   }
 
   public MarkableInputStream(InputStream in, int size) {
+    this(in, size, DEFAULT_LIMIT_INCREMENT);
+  }
+
+  private MarkableInputStream(InputStream in, int size, int limitIncrement) {
     if (!in.markSupported()) {
       in = new BufferedInputStream(in, size);
     }
     this.in = in;
+    this.limitIncrement = limitIncrement;
   }
 
   /** Marks this place in the stream so we can reset back to it later. */
@@ -64,6 +71,9 @@ final class MarkableInputStream extends InputStream {
     return offset;
   }
 
+  public void allowMarksToExpire(boolean allowExpire) {
+      this.allowExpire = allowExpire;
+  }
   /**
    * Makes sure that the underlying stream can backtrack the full range from
    * {@code reset} thru {@code limit}. Since we can't call {@code mark()}
@@ -119,6 +129,9 @@ final class MarkableInputStream extends InputStream {
   }
 
   @Override public int read() throws IOException {
+    if (!allowExpire && (offset + 1 > limit)) {
+      setLimit(limit + limitIncrement);
+    }
     int result = in.read();
     if (result != -1) {
       offset++;
@@ -127,6 +140,9 @@ final class MarkableInputStream extends InputStream {
   }
 
   @Override public int read(byte[] buffer) throws IOException {
+    if (!allowExpire && (offset + buffer.length > limit)) {
+      setLimit(offset + buffer.length + limitIncrement);
+    }
     int count = in.read(buffer);
     if (count != -1) {
       offset += count;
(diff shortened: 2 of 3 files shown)


picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
```diff
@@ -119,10 +119,10 @@
    * {@code inSampleSize}).
    */
   static Bitmap decodeStream(InputStream stream, Request request) throws IOException {
-    MarkableInputStream markStream = new MarkableInputStream(stream);
+    MarkableInputStream markStream = new MarkableInputStream(stream, true);
     stream = markStream;
 
-    long mark = markStream.savePosition(65536); // TODO fix this crap.
+    long mark = markStream.savePosition(65536);
```
JakeWharton: This value should be 1024

picasso/src/main/java/com/squareup/picasso/MarkableInputStream.java
```diff
@@ -33,16 +33,26 @@
   private long reset;
   private long limit;
   private long defaultMark = -1;
+  private boolean dynamicLimitEnabled = false;
```
JakeWharton: I have no interest in making this class generic. Eliminate this boolean and always perform this behavior.

picasso/src/test/java/com/squareup/picasso/MarkableInputStreamTest.java
```diff
@@ -52,6 +52,35 @@ public void test() throws Exception {
...
+    in.reset(); // Back to 'H'
+    assertThat(readBytes(in, 3)).isEqualTo("HIJ");
+    in.reset(posA); // Back to 'D'
+    assertThat(readBytes(in, 7)).isEqualTo("DEFGHIJ");
+    in.reset(); // Back to 'H' again.
+    assertThat(readBytes(in, 6)).isEqualTo("HIJKLM");
+    try {
+      in.reset(); // Back to 'H' again despite the limit being exceeded
+      assertThat(readBytes(in, 2)).isEqualTo("HI"); // confirm we're really back at H
+    } catch (IOException notexpected) {
```
JakeWharton: You can delete both of these try/catch blocks. If an exception is thrown it'll fail the test.

picasso/src/test/java/com/squareup/picasso/MarkableInputStreamTest.java
```diff
@@ -40,16 +40,33 @@ public void test() throws Exception {
...
+    assertThat(readBytes(in, 7)).isEqualTo("DEFGHIJ");
+    in.reset(); // Back to 'H' again.
+    assertThat(readBytes(in, 6)).isEqualTo("HIJKLM");
+    
+    in.reset(); // Back to 'H' again despite the original limit being exceeded
+    assertThat(readBytes(in, 2)).isEqualTo("HI"); // confirm we're really back at H
+
+    in.reset(posA); // Back to 'D' again despite the original limit being exceeded
+    assertThat(readBytes(in, 2)).isEqualTo("DE"); // confirm that we're really back at D
+
```
dnkoutso: nit: extra line.

dnkoutso:
Have you attempted to load some of the images in the issues reported? I will do this for you now.

dnkoutso:
Also please squash to one commit.

dnkoutso:
Confirmed that this now works with

`http://files.parsetfss.com/8f840861-7654-48ce-a603-b1e7f4fa1e25/tfss-f6a755f2-2a8f-4ba5-bea0-2ae895294dcb-FlyerApp.png`

picasso/src/main/java/com/squareup/picasso/MarkableInputStream.java
```diff
@@ -127,6 +131,9 @@ private void skip(long current, long target) throws IOException {
   }
 
   @Override public int read(byte[] buffer) throws IOException {
+    if (offset + buffer.length >= limit) {
+      setLimit(2 * (offset + buffer.length));
```
JakeWharton: So this behavior isn't quite right. We need the buffering of data to only happen on when a marker is set (for reading the image metadata) but to NOT happen when there is no marker (during the actual full image decode). As it stands now this will cause a ton of allocation and GC churn in the latter case when we don't actually care about buffering.

My approach was going to be killing this class for another `InputStream` decorator that used an on-demand list of 2K `byte[]` chunks for infinite buffering when a mark was set. In the ideal case only a single 2K chunk would be allocated. In the pathological degradation case it would just allocate additional chunks as needed until reset.

JohnWowUs: I see the problem. I think the solution would be to let the mark(s) to expire once there has been a reset back to the earliest mark.

JakeWharton: Yeah. If we can detect that case and not change the limit then I think this will work.

JohnWowUs:
Tried a number of "clever" solutions but they all had flaws so I went with something simple. The default behaviour is to allow the marks to expire. If, however, a flag is set than the read limit will increase when needed.

start141:
This code can't load pictures which exifOrientation == 90 or 270.

Please test #584 @JohnWowUsThis, thanks!

JohnWowUs:
@start141 
The  EXIF changes committed to the master only handle EXIF orientations for local images. There's a PR to extend this to network images but it's still being evaluated. Could you try with local images and post the stack trace?

start141:
@JohnWowUs 
It is also can't load local images which exif orientation 90 or 270.

start141:
Found a picture can't load, and it's orientation isn't 90 or 270.
56 1ca3c34c8e9aa5 297x222

start141:
When I debug,
I write
  markStream.allowMarksToExpire(false);
  mark = markStream.savePosition(1024);
to
 markStream.allowMarksToExpire(false);
 mark = markStream.savePosition(65536);

Finally, I found a interesting thing, it's load all images which I say.

What about mean？

JohnWowUs:
It's not a problem with the rotation. There's a problem with how the readlimit is recalculated in certain cases in this PR.

start141:
Thank you!
So I can use savePosition(65536) provisional? any bad things will happen？

JohnWowUs:
It shouldn't do any harm (other than some images not loading) but you might be better off using the current picasso master branch and increasing the marker limit to something like 131072.

JohnWowUs:
The bug should be fixed in the latest commit.

JakeWharton:
Will look soon. Thanks.

tvietinghoff:
I experienced the same issue on Nexus 9 / 5.1.1 and Galaxy S4 4.4.2. I can confirm that the changes from this pull request fixed the issue for me. Please roll this into an official build soon.

damianw:
@JohnWowUs's fix appears to have fixed the issue for me. I haven't seen any new reports come in since applying this patch.

UglyTroLL:
Any updates on this fix?

cbruegg:
After several days of testing, I too can confirm that this has fixed the issue.

picasso/src/main/java/com/squareup/picasso/MarkableInputStream.java
```diff
@@ -119,6 +129,9 @@ private void skip(long current, long target) throws IOException {
   }
 
   @Override public int read() throws IOException {
+    if (!allowExpire && (offset + 1 > limit)) {
+      setLimit(limit + limitIncrement);
```
JakeWharton: Ideally this limit doubles every time its hit. Otherwise you needlessly churn through memory and waste copies in the pathological degradation case that this is meant to fix.

rogerhu:
friendly ping here -- 2.5.3-SNAPSHOT fixes the large image file loading issue.  Wondering if there's a timeline to release 2.5.3?

JakeWharton:
There is not, no.

Issue #1044, opened 2015-05-31
Picasso leaks context even though cancel is called


mblcdr:
I load my photo using:
```Java
Picasso.with(this).load(imageUrl)
                .tag(PhotoActivity.class)
                .resize(w, h)
                .into(photo, new Callback() {
                         @Override
                         public void onSuccess() {
                              //Some method calls here
                         }

                         @Override
                         public void onError() {}
                });
```
And I cancel this request in ```onStop()``` method using this line of code:
```Java
Picasso.with(this).cancelTag(PhotoActivity.class);
```

Yet I get this log from leak canary:

    * com.example.app.activity.PhotoActivity has leaked:
    * GC ROOT thread com.squareup.picasso.Utils$PicassoThread.<Java Local> (named 'Picasso-Idle')
    * references java.util.concurrent.PriorityBlockingQueue.queue
    * references array java.lang.Object[].[0]
    * references com.squareup.picasso.PicassoExecutorService$PicassoFutureTask.hunter
    * references com.squareup.picasso.BitmapHunter.action
    * references com.squareup.picasso.ImageViewAction.callback
    * references com.example.app.activity.PhotoActivity$1.this$0 (anonymous class implements com.squareup.picasso.Callback)
    * leaks com.example.app.activity.PhotoActivity instance
    * Reference Key: 0a9592c7-a4b5-4ff0-adf9-d3fa9984a71a
    * Device: Sony Sony C2105 C2105_1271-3303
    * Android Version: 4.2.2 API: 17
    * Durations: watch=5006ms, gc=181ms, heap dump=808ms, analysis=24507ms

mblcdr:
I guess there was a problem with my setup, using Picasso multiple times. After I make sure every request has the same tag the leak was gone.
Sorry for the spam.

mblcdr:
The leak happened again. All of my Picasso calls that have callbacks use PhotoActivity.class as tag, but the leak is happening again.

dnkoutso:
Found one. And I am testing yours now. PR coming up.

dnkoutso:
I cant seem to repro your exact case. I did find another potential leak with `fit()` though and I fixed that.

dnkoutso:
```java    // Trigger the download of the URL asynchronously into the image view.
    Picasso.with(context) //
        .load(url) //
        .placeholder(R.drawable.placeholder) //
        .error(R.drawable.error) //
        .tag(SampleGridViewActivity.class) //
        .into(view, new Callback() {
[...]

PR #1046, opened 2015-05-31
Fix potential leak with deferred request creators

dnkoutso:
If for any reason a `cancelTag` is invoked before a `DeferredRequestCreator` has executed the request creator tag will leak temporarily until its get flushed out. This is bad if the tag is an activity or fragment.

found from #1044


JakeWharton:
Needs rebase. Otherwise, LGTM merge away.

PR #1067, opened 2015-06-17
When center cropping, add a top crop option for portrait bitmaps.

Vilgan:
This request is something that initially I thought was a small/irrelevant request but has come up multiple times. When cropping bitmaps, various people have requested the ability to crop focusing on the top for portrait images rather than focusing on the center in order to reduce the risk of a head or face being cut off.

The actual code is super simple, but making it fit cleanly into existing options is less obvious. It seemed better to treat this as an extension to center crop because the behavior is basically the exact same but with one line changed to set drawY to 0 instead of splitting the offset in some cases.

If there is a better way for this to fit in, please feel free to call it out and I'll adjust the PR.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -577,13 +577,16 @@ class BitmapHunter implements Runnable {
         float heightRatio =
             targetHeight != 0 ? targetHeight / (float) inHeight : targetWidth / (float) inWidth;
         float scaleX, scaleY;
+        boolean shouldTopCrop = data.topCropPortrait && (inHeight > inWidth);
         if (widthRatio > heightRatio) {
+          // Need to scale and some vertical portions will not be visible
           int newSize = (int) Math.ceil(inHeight * (heightRatio / widthRatio));
-          drawY = (inHeight - newSize) / 2;
+          drawY = shouldTopCrop ? 0 : (inHeight - newSize) / 2;
           drawHeight = newSize;
           scaleX = widthRatio;
           scaleY = targetHeight / (float) drawHeight;
         } else if (widthRatio < heightRatio) {
+          // Need to scale and some horizontal portions will not be visible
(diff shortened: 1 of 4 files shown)


JakeWharton:
I would rather use arguments of `Gravity.TOP`, `Gravity.BOTTOM`, `Gravity.LEFT`, `Gravity.RIGHT`, and `Gravity.CENTER` for this behavior to a `crop()` method. `centerCrop()` would become an alias for `crop(CENTER)`.

Who's to say that other people don't have use cases where other edges are more favorable include beyond just top?

Vilgan:
That makes sense, I'll drop this PR and do something like that. How do I understand what check style errors are being thrown by Travis CI? The log refers to them but doesn't seem to call out what they are specifically.

JakeWharton:
Scroll up a bit
```
/home/travis/build/square/picasso/picasso/src/main/java/com/squareup/picasso/Request.java:98: Line is longer than 100 characters (found 107).
/home/travis/build/square/picasso/picasso/src/main/java/com/squareup/picasso/Request.java:364: Line is longer than 100 characters (found 103).
/home/travis/build/square/picasso/picasso/src/main/java/com/squareup/picasso/Request.java:365: Line is longer than 100 characters (found 101).
/home/travis/build/square/picasso/picasso/src/main/java/com/squareup/picasso/Request.java:366: Line is longer than 100 characters (found 102).
/home/travis/build/square/picasso/picasso/src/main/java/com/squareup/picasso/Request.java:513: Line is longer than 100 characters (found 116).
/home/travis/build/square/picasso/picasso/src/main/java/com/squareup/picasso/RequestCreator.java:247: Line is longer than 100 characters (found 101).
```

seshuvinay:
I am actually using a library to find out coordinates of face in an image. Is there a way I can use this library to display image by cropping it so that face is included. 

Eg. If it is a portrait image with face at the bottom of image it should be cropped from the bottom. If face is equidistant from top and bottom, it should be center cropped and if it's at the top, it should be top cropped.

PR #1080, opened 2015-07-12
Naturalize's suggestions

condda:
Naturalize's renaming suggestions to improve the codebase consistency

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso/Request.java
@@ -113,45 +113,45 @@ public final class Request {
   }
 
   @Override public String toString() {
-    final StringBuilder sb = new StringBuilder("Request{");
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -509,11 +509,11 @@ Bitmap quickMemoryCacheCheck(String key) {
   }
 
   void complete(BitmapHunter hunter) {
-    Action single = hunter.getAction();
+    Action action = hunter.getAction();
```
JakeWharton: This should remain as "single". Hunters have a "single" action and also "joined" actions.

condda:
Ok, I'll remove that commit

JakeWharton:
Thanks!

condda:
My pleasure

PR #1090, opened 2015-07-24
Stop placeholder animation on errors

jobi:


JakeWharton:
Are you able to write a test for this? Look at the test which verifies we call `start()` for ideas, perhaps.

jobi:
Added a test. By the way I couldn't find the test that checks that start() is called?

JakeWharton:
Oh maybe I lied! Will look later.

JakeWharton:
Thanks!

PR #1120, opened 2015-08-20
shouldResize with zero target

PeterBergmann:
When onlyScaleDown is true and one of the target dimensions are set to 0 it will resize bitmap despite other dimension is smaller than target.


mblcdr:
@JakeWharton This could be closed now that #1456 is merged.

JakeWharton:
Thanks!

PR #1130, opened 2015-09-04
Use Gradle for build system #876, allow building against higher android SDK

mullender:
This change migrates from maven based builds to gradle based builds.

The gradle build produces:
- source and javadoc jars
- aar libraries for picasso and picasso-pollexor

./gradlew uploadArchives has not been tested, but the artifacts have been verified.


dnkoutso:
dup #876

mullender:
Not a dup just adding the things that were missing. Like the maven publishing. 

I need it to add bitmap pooling.

JakeWharton:
#1409

PR #1134, opened 2015-09-08
Optimization image cache

recoilme:
- 2Q: A Low Overhead High Performance Buffer Management Replacement Algorith
- Based on description: http://www.vldb.org/conf/1994/P439.PDF
- Created by recoilme on 22/08/15.
- email: vadim-kulibaba@yandex.ru

Hi guys.
LRU cache is simple & fast but sucks in real world app, especially for bitmaps. Medium & large sized bitmaps evict small images. I research caching alghoritms and chhose 2 queue algorithm as simple and fast and realize it. Now i have very good hitrate for repeating images, and surprize - it's faster then LRU. I implement it for picasso and write some tests. I will be happy see it in next release. Fill free send me email if you have questions.

Thanks for your work, Vadim

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -853,7 +853,8 @@ public class Picasso {
         downloader = Utils.createDefaultDownloader(context);
       }
       if (cache == null) {
-        cache = new LruCache(context);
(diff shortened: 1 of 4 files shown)


recoilme:
Tests not passed because tests was changed? Must i fix it?

JakeWharton:
I'd be okay changing the cache on master to something more efficient, but I don't want a second cache implementation.

PR #1146, opened 2015-09-21
Add support to dither a decoded image.

miksto:
Add support to dither decoded images. Using the bitmap config RGB\_565 causes banding issues in the image, which dithering will make less apparent.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso/Request.java
@@ -83,13 +83,16 @@ public final class Request {
   public final boolean purgeable;
   /** Target image config for decoding. */
   public final Bitmap.Config config;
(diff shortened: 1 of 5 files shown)


kwangsu61:
@JakeWharton I need inDither and inPreferQualityOverSpeed options.
- inDither : in case of RGB_565, I need for better image quality
- inPreferQualityOverSpeed : on some Samsung's devices, I need for better image quality
https://developer.android.com/reference/android/graphics/BitmapFactory.Options.html

Do you have to plan to add this options?

JakeWharton:
Do you want control per image or just setting them for an entire Picasso
instance?

kwangsu61:
@JakeWharton we need control per image request like purgeable(), because we want to use this options in only some cases. Thanks!

PR #1149, opened 2015-09-24
Count size of loops once instead of each iteration

AlmightyMegadeth00:
As suggested here: http://developer.android.com/training/articles/perf-tips.html


JakeWharton:
All of these loops already use this technique.

Issue #1191, opened 2015-11-06
Images don't download if url redirects


Blahartinger:
If I give it an image url with http://... and that happens to redirect to https://... then the image always fails to load.

jonan:
Redirects are handled by the HTTP client, not Picasso. You should file a bug against the HTTP client that's giving you problems.

Blahartinger:
But Picasso still calls its onsuccess method in its callback, is that not weird?

Also, perhaps there is something to configure on the default http client that was missed.

JakeWharton:
I cannot reproduce this with HttpURLConnection. The client doesn't follow redirects, but it also doesn't result in onSuccess being called. Please provide a test or sample that showcases the problem.

Blahartinger:
HttpURLConnection doesn't follow redirects automatically, true...but the Picasso class, UrlConnectionDownloader that uses it, should handle this. 

Often a 300 is indicative of redirect, but UrlConnectionDownloader just disconnects the request.

int responseCode = connection.getResponseCode();
    if (responseCode >= 300) {
      connection.disconnect();
      throw new ResponseException(responseCode + " " + connection.getResponseMessage(),
          networkPolicy, responseCode);
    }

Should it not try the redirected url?

 int status = connection.getResponseCode();
                if (status != HttpURLConnection.HTTP_OK) {
                    if (status == HttpURLConnection.HTTP_MOVED_TEMP
                            || status == HttpURLConnection.HTTP_MOVED_PERM
                            || status == HttpURLConnection.HTTP_SEE_OTHER) {
                        // Redirect

                        // get redirect url from "location" header field
                        String newUrl = connection.getHeaderField("Location");

                        // open the new connection
                        connection = (HttpURLConnection) new URL(newUrl).openConnection();

                    }
                }

Blahartinger:
I've created a PR for this change if that seems appropriate to you guys.
https://github.com/square/picasso/pull/1192

abdul-martinez:
Is there an estimation of when this PR will be merged?

lkesteloot:
Many of my users were affected today because of this bug in UrlConnectionDownloader. Can you merge the PR please? Note that the PR is not complete: It should follow multiple (but finite) redirects and it should close the connection. I'll make my own PR if you're willing to merge it.

PR #1192, opened 2015-11-06
Handle URL Redirects.

Blahartinger:
To fix issue https://github.com/square/picasso/issues/1191

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/UrlConnectionDownloader.java
+++ b/picasso/src/main/java/com/squareup/picasso/UrlConnectionDownloader.java
@@ -91,10 +91,23 @@ public class UrlConnectionDownloader implements Downloader {
     }
 
     int responseCode = connection.getResponseCode();
-    if (responseCode >= 300) {
-      connection.disconnect();
-      throw new ResponseException(responseCode + " " + connection.getResponseMessage(),
-          networkPolicy, responseCode);
+    if (responseCode != HttpURLConnection.HTTP_OK) {
+      if (responseCode == HttpURLConnection.HTTP_MOVED_TEMP
+              || responseCode == HttpURLConnection.HTTP_MOVED_PERM
+              || responseCode == HttpURLConnection.HTTP_SEE_OTHER) {
+        // Redirect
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso/UrlConnectionDownloader.java
```diff
@@ -91,10 +91,23 @@ protected HttpURLConnection openConnection(Uri path) throws IOException {
...
-      throw new ResponseException(responseCode + " " + connection.getResponseMessage(),
-          networkPolicy, responseCode);
+    if (responseCode != HttpURLConnection.HTTP_OK) {
+      if (responseCode == HttpURLConnection.HTTP_MOVED_TEMP
+              || responseCode == HttpURLConnection.HTTP_MOVED_PERM
+              || responseCode == HttpURLConnection.HTTP_SEE_OTHER) {
+        // Redirect
+
+        // get redirect url from "location" header field
+        String newUrl = connection.getHeaderField("Location");
```
Jawnnypoo: Shouldn't we first call `connection.disconnect()` before reassigning this? Docs state "Releases this connection so that its resources may be either reused or closed" so it sounds like we would want to call `disconnect` before reassignment

picasso/src/main/java/com/squareup/picasso/UrlConnectionDownloader.java
```diff
@@ -91,10 +91,23 @@ protected HttpURLConnection openConnection(Uri path) throws IOException {
...
+
+        // open the new connnection again
+        connection = (HttpURLConnection) new URL(newUrl).openConnection();
```
Jawnnypoo: Shouldn't we first call connection.disconnect() before reassigning this? Docs state "Releases this connection so that its resources may be either reused or closed" so it sounds like we would want to call disconnect before reassignment

dnkoutso:
Can you add a test?

JakeWharton:
We are removing support for HttpURLConnection in the next major version so this is no longer needed.

PR #1206, opened 2015-11-18
Catch OutOfMemoryError caused when setting error image

jonathan-caryl:
I wrote a simple app to reproduce the crash we occasionally were seeing via Crashlytics. 

https://github.com/jonathan-caryl/PicassoOnErrorOOMTest

It seems better to swallow the OOM and not display an error image than to crash.


JakeWharton:
This just defers the problem and the app will crash somewhere else. OOMs are rarely stem from loading images, but rather because they are so large images end up triggering the OOMs. If your app is putting that much pressure on the heap it's indicative of a larger problem.

Besides, this violates the contract of Picasso. If you ask for an error image we are going to load one.

PR #1218, opened 2015-11-30
BitmapHunter.decodeStream uses larger read limit 

purdyk:
BitmapHunter.decodeStream uses more reasonable read limit when saving position.  Left //TODO in because this seems likely to continue to cause problems in the future.

A change somewhere into 2.5 has created issues in this area.  Our app attempts to load some thumbnails from disk, but fails at L140 as parsing the bitmap header on some devices has resulted in reading past 65535 bytes.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -114,7 +114,7 @@ class BitmapHunter implements Runnable {
     MarkableInputStream markStream = new MarkableInputStream(stream);
     stream = markStream;
 
-    long mark = markStream.savePosition(65536); // TODO fix this crap.
+    long mark = markStream.savePosition(67108863); // TODO fix this crap.
 
     final BitmapFactory.Options options = RequestHandler.createBitmapOptions(request);
     final boolean calculateSize = RequestHandler.requiresInSampleSize(options);


purdyk:
I considered simply using Integer.MAX_VALUE instead of the continuing the arbitrary use of large numbers, but I settled on another large value ( 2^26 -1 ) around 70mb because perhaps there is a problem with an image with 70mb of headers.

JakeWharton:
This is not the right fix, and will cause a needless allocation at the specified size for _all_ images.

purdyk:
MarkableInputStream doesn't perform any allocations.  This change has no memory impact whatsoever.  It does allow photos taken by my Moto X to be loaded by picasso. 
After the first pass on my device, the call to BitmapFactory.decodeStream with options.inJustDecodeBounds set to true, offset is 739114 for a sample image.
I'd love a recommendation for an alternative fix for this, but I do not see a possibility other than simply using a smaller magic number.

JakeWharton:
The contained `BufferedInputStream` allocates a `byte[]` for the given size.

JakeWharton:
And the fix is to just use a growable buffer instead of a fixed one.

purdyk:
Jake thanks for taking the time to review and point out my error.
Before I start an implemntation that I will be submitting, I'd like to know if you have any objection to adding the buffer growing functionality to MarkableInputStream?

PR #1219, opened 2015-11-30
MarkableInputStream automatic buffer growth.

purdyk:
MarkableInputStream now has the ability to automatically grow the underlying input stream's buffer.
BitmapHunter uses this ability to ensure that images which require reading more than 2048 bytes
 to determine bounds succeed in loading.

Buffer grows by either required projected limit or the current limit, whichever is larger.


JakeWharton:
I'll look tomorrow, but the default should be 2048. Almost all headers are
way before this.

UglyTroLL:
There is a PR for this issue https://github.com/square/picasso/pull/1035

JakeWharton:
A fix was merged.

PR #1225, opened 2015-12-07
Parametrised crop using Gravity flags

jskierbi:
I've added customisable crop() with android.view.Gravity flags as parameters.

This change enables to call such code:

```java
Picasso.with(context)
        .load(imageUrl)
        .fit()
        .crop(Gravity.START | Gravity.BOTTOM)
        .into(imageView);
```

This will make Picasso position image first and then perform crop. For example above, image start and bottom will be shown (cropping end or top part of the image)

Code is backwards compatible, centerCrop() call is now alias to crop(Gravity.CENTER).


marcinbak:
Perfect. Just something I need in my project!

maciekczwa:
+1

jskierbi:
I'm changing API a bit and will open another PR with same feature.

Issue #954, opened 2015-03-28
GridLayoutManager - wrong image loaded when notifyItemChanged() is called.


gswierczynski:
I am not sure if this is Picasso issue, GridLayoutManager issue or my implementation is just flawed (see the sample app).

The issue is described on stackoverflow under:
http://stackoverflow.com/questions/29311428/recyclerview-with-gridlayoutmanager-and-picasso-showing-wrong-image

### Sample app:
git clone https://github.com/gswierczynski/recycler-view-grid-layout-with-picasso.git
Tap on an item calls notifyItemChanged with parameter equal to the position of that item.

### Reproduction:

RecyclerView with GridLayoutManager (spanCount = 3). List items are CardViews with ImageView inside.

When all the items does not fit the screen calling notifyItemChanged on one item causes more than one calls to onBindViewHolder(). One call is for position from notifyItemChanged others for items not visible on the screen.

### Issue:

Sometimes the item at position passed to the notiryItemChanged is loaded with an image belonging to an item that is not on the screen (most likely due to recycling of the view holder - although I would assume that if the item remains in place then the passed viewholder would be the same).

I have found Jake's comment on other issue here about calling load() even if the file/uri is null. Image is loaded on every onBindViewHolder here.

JakeWharton:
Thanks. If only all issues reported would be as well executed...

We'll have a look soon!

gswierczynski:
Any luck reproducing it on your end?

gswierczynski:
I think the issue is connected in some way to the behavior of ItemAnimator.
Please see the explanation [HERE](https://code.google.com/p/android/issues/detail?id=162699) the 
When setSupportsChangeAnimations(false) is used the issue is not present.

dnkoutso:
Thanks for the very detailed bug report. I havent read the whole thing but did you come into a conclusion to this?

fahimk:
I had the same issue in my app where calling notifyItemChanged(position) flashed the item and an image from one of the other tiles would replace the one on the item that was notified.

setSupportsChangeAnimations(false) and setHasStableIds(true) seem to prevent this from happening

losogithub:
Same issue.
@gswierczynski 's report is quite enough.
Looking forward for a fix.
Thx a lot.

erg:
What's the status on merging this bugfix?

JakeWharton:
6991476576985d5d1cec1b37b61bc24d1d4bf4c7

p-fischer:
[...]

PR #1228, opened 2015-12-08
Fix wrong image loaded when notifyItemChanged() is called

losogithub:
https://github.com/square/picasso/issues/954
cause: 
  onPreDraw() is called several times even after removeOnPreDrawListener() has been called.
precondition:
  1 using RecyclerView
  2 calling fit()
  3 calling notifyItemChanged() or notifyItemRangeChanged()
solution:
  Clear target reference after removeOnPreDrawListener to ensure not calling into() again.


JakeWharton:
Added a test and merged as 6991476576985d5d1cec1b37b61bc24d1d4bf4c7

PR #1250, opened 2016-01-02
The fix of unnecessary objects creation

agap:
The `getTrianglePath()` method creates two `Point` objects which will be garbage-collected in no time. These objects can be safely replaced with plain `int` variables to save some work of garbage collector (which conforms to Effective Java item №5 - "Avoid creating unnecessary objects").

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
@@ -21,7 +21,6 @@ import android.graphics.Canvas;
 import android.graphics.ColorFilter;
 import android.graphics.Paint;
 import android.graphics.Path;
-import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.AnimationDrawable;
 import android.graphics.drawable.BitmapDrawable;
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
```diff
@@ -153,13 +153,13 @@ private void drawDebugIndicator(Canvas canvas) {
   }
 
   private static Path getTrianglePath(Point p1, int width) {
```
JakeWharton: How about getting rid of this `Point` as well?

agap: @JakeWharton yeah, sure. Is it OK if I fix it in another commit, or it will be more correct to create a fresh branch, fix the `getTrianglePath()` signature, its implementation and usage in one commit and make another pull-request?

JakeWharton: You can amend the commit and force push your branch. It will automatically update this pull request.

agap:
That's strange. All tests performed on JDK 7 were positive, but `NullPointerException` in `Picasso-Dispatcher` thread is thrown again and again, failing the whole build.

JakeWharton:
I'll look tomorrow.

agap:
Hi, @JakeWharton. Was the build error somehow caused by my changes or the issue was somewhere else?

PR #1252, opened 2016-01-06
add sbt dependency to readme

joprice:


hzsweers:
I don't think this is necessary

JakeWharton:
I'm inclined to agree since we'll be removing Maven soon.

PR #1257, opened 2016-01-10
Add a feature to globally disable fade

robertdolca:
This is useful when testing the app. Looks like Spoon sometimes takes the screenshot while the fade is running.


dnkoutso:
Needs tests.

JakeWharton:
Duping onto #1356 for a general solution.

PR #1262, opened 2016-01-12
No retry on FileNotFound exception

bwill:
The existing code will queue up a retry (with a 500ms delay),
and then ask the RequestHandler if retry is allowed.
For most handlers (e.g. AssetRequestHandler and FileRequestHandler),
it'll return false, and queue up a fail event.
This change will skip directly to fail, so the callback will be
triggered much faster.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -19,6 +19,8 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Matrix;
 import android.net.NetworkInfo;
+
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


JakeWharton:
So to summarize, rather than retrying 3 times this will immediately fail the request?

bwill:
The existing code path doesn't retry 3 times in this case.  It calls dispatchRetry, and in performRetry, it calls `hunter.shouldRetry`, which returns false, and then it calls `performError`, without ever "re-hunting".

With the existing behavior, it makes loads for missing files seem slow, because the error result is delayed 500ms (+ up to 200 ms for the batch delay for the callback event).

bwill:
My assumption was that a networked request wouldn't return FileNotFoundException (but rather a plain IOException or some other subtype), but I just checked and discovered that's not true, so I'm closing this.

In principle, are you opposed to having local request handlers (like File, Asset and Resource) fail fast?

JakeWharton:
I am on board with the desired behavior, yes.

PR #1270, opened 2016-01-18
Support removing items from the memory cache by resource id.

lightai:


JakeWharton:
When would a resource change while the app was running?

lightai:
@JakeWharton 
Activity#onCreate() {
		// pre loading a big image (by resource id)
		Picasso.with(this).load(R.drawable.asset_mode_tile).fetch();
}

Activity#onDestroy() {
		// clear memory of pre loaded big image (but by path?)
		Picasso.with(this).invalidate(String.valueOf(R.drawable.asset_mode_tile));
}

better if have a invalidate by resource id api.

JakeWharton:
You can use `.cachePolicy(NO_STORE)` on the initial request to not put the image in the memory cache which will have the same effect.

PR #1296, opened 2016-02-13
Clear the reference to the placeholder when not needed anymore

### LINKED ISSUE #968: PicassoDrawable's placeholder is not released if the drawable becomes another PicassoDrawable's placeholder.
I have an ImageView that loops through remote images like a "slideshow".

Because I still want to display the last image while it's loading from the server, I do `setPlaceholder(target.getDrawable())` which sets the previous PicassoDrawable as the placeholder of the new PicassoDrawable.

This leaks memory over time because the every PicassoDrawable is holding a reference to the previous PicassoDrawable through `placeholder` member variable.

GuillaumeDelente:
Should fix #968

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoDrawable.java
@@ -92,6 +92,7 @@ final class PicassoDrawable extends BitmapDrawable {
 
   @Override public void draw(Canvas canvas) {
     if (!animating) {
+      placeholder = null;
       super.draw(canvas);
     } else {
       float normalized = (SystemClock.uptimeMillis() - startTimeMillis) / FADE_DURATION;


JakeWharton:
Thanks.

JakeWharton:
The placeholder image is set to `null` when the animation completes and if you are not animating then the placeholder field is never set. So I'm not sure this actually fixes anything.

JakeWharton:
I'm reverting this and re-opening the original issue because I'm not confident this fixes any problem.

GuillaumeDelente:
I remember the placeholder being set even if not animating but need to confirm.

JakeWharton:
The constructor contains this code:
```java
     boolean fade = loadedFrom != MEMORY && !noFade;
     if (fade) {
       this.placeholder = placeholder;
       animating = true;
       startTimeMillis = SystemClock.uptimeMillis();
```
but I'd definitely be interested in figuring out what's going on here.

PR #1313, opened 2016-03-11
Update SampleGalleryActivity.java

shuhuaxie:
Fix bug found in my phone that picture is too large to load.

### DIFF
--- a/picasso-sample/src/main/java/com/example/picasso/SampleGalleryActivity.java
+++ b/picasso-sample/src/main/java/com/example/picasso/SampleGalleryActivity.java
@@ -71,7 +71,7 @@ public class SampleGalleryActivity extends PicassoSampleActivity {
     // Index 1 is the progress bar. Show it while we're loading the image.
     animator.setDisplayedChild(1);
 
-    Picasso.with(this).load(image).into(imageView, new EmptyCallback() {
+    Picasso.with(this).load(image).fit().into(imageView, new EmptyCallback() {
       @Override public void onSuccess() {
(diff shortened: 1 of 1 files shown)


picasso-sample/src/main/java/com/example/picasso/SampleGalleryActivity.java
```diff
@@ -71,7 +71,7 @@ private void loadImage() {
     // Index 1 is the progress bar. Show it while we're loading the image.
     animator.setDisplayedChild(1);
 
-    Picasso.with(this).load(image).into(imageView, new EmptyCallback() {
+    Picasso.with(this).load(image).fit().into(imageView, new EmptyCallback() {
```
JakeWharton: You probably also want `.centerInside()` to maintain aspect ratio.

shuhuaxie: It will be great to add ".centerInside()",but I think ".fit()" is enough to deal my problem.Thanks.

JakeWharton: Well currently you'll be distorting the aspect ratio with just `.fit()` so I'm not going to merge it unless you add `.centerInside()`.

shuhuaxie: Ok, I add ".centerInside()" and make a test of it, It looks very good.

JakeWharton:
Squashed and merged. Thanks.

PR #1322, opened 2016-03-22
remove cache

selabku:
remove picasso all cache files


JakeWharton:
This class isn't public so adding this method won't be very useful.

Why do you want this?

selabku:
I want to update some image periodically, but there is no method to delete cache.

PR #1325, opened 2016-03-22
Deferred requests on unattached views should defer VTO listener until…

### LINKED ISSUE #1321: ViewTreeObserver API is mis-used
We've received the following bug about RecyclerView and Picasso:
https://code.google.com/p/android/issues/detail?id=191824

After investigation, it turns out that it is an issue related to how ViewTreeObservers work. Basically, two  `view#getViewTreeObserver` calls wont necessarily return the same observer if the View's attach/detach state changes in between. 

When you add an observer to a View while it is detached, you actually add it to an internal tree and the list gets merged into the main ViewTree once the view is attached. Unfortunately, they are NOT detached when the view is detached. 
So there is a chance that the observer will never be removed because you'll keep removing it from the fake one here:
https://github.com/square/picasso/blob/bc6af7a755a30e98be4cf57cb77f36809ddb92ba/picasso/src/main/java/com/squareup/picasso/DeferredRequestCreator.java#L40

Although the behavior in the Framework is not intuitive, there is not much we can do at this point since changing it will mean behavior change and we cannot change existing devices. 

You can work around this issue by adding an attach state listener to the View.

JakeWharton:
… attach.

Closes #1321.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/DeferredRequestCreator.java
+++ b/picasso/src/main/java/com/squareup/picasso/DeferredRequestCreator.java
@@ -15,15 +15,21 @@
  */
 package com.squareup.picasso;
 
+import android.view.View;
+import android.view.View.OnAttachStateChangeListener;
 import android.view.ViewTreeObserver;
 import android.widget.ImageView;
 import java.lang.ref.WeakReference;
 import org.jetbrains.annotations.TestOnly;
 
+import static android.os.Build.VERSION.SDK_INT;
+import static android.os.Build.VERSION_CODES.HONEYCOMB_MR1;
+
 class DeferredRequestCreator implements ViewTreeObserver.OnPreDrawListener {
 
   final RequestCreator creator;
   final WeakReference<ImageView> target;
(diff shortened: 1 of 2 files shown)


dnkoutso:
:clap:

cxzhang2:
@JakeWharton unless I'm misunderstanding #1321 (very possible), I don't think this PR solves the issue completely. 

Scenario: regardless of when we addPreDrawListener, once the target view becomes attached-detached again, calling target.getVTO() will return the fake tree while our listener is attached to the real tree, and calling remove() on the fake tree won't remove our listener from the real tree.

I think the correct fix is to defer all .getViewTreeObserver() calls until attachment.

JakeWharton:
Yep I made that fix locally but never got around to writing tests for it and sending a PR.

cxzhang2:
Awesome, I'm not crazy

JakeWharton:
Heh, nope. I've been trying to circle back to Picasso but it's hard to
justify the time working on it.

PR #1323, opened 2016-03-22
Set max pool size before core to avoid IAE by OpenJDK implementation.

JakeWharton:
Closes #1319

Issue #1319, opened 2016-03-16
Android N: IllegalArgumentException


phileo:
I have  2 crash reports from out in the field, both on Android N (Nexus 6P) devices:

```
Fatal Exception: java.lang.IllegalArgumentException
       at java.util.concurrent.ThreadPoolExecutor.setCorePoolSize(ThreadPoolExecutor.java:1529)
       at com.squareup.picasso.PicassoExecutorService.setThreadCount(SourceFile:81)
       at com.squareup.picasso.PicassoExecutorService.adjustThreadCount(SourceFile:51)
       at com.squareup.picasso.Dispatcher.performNetworkStateChange(SourceFile:395)
       at com.squareup.picasso.Dispatcher$DispatcherHandler.handleMessage(SourceFile:515)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:148)
       at android.os.HandlerThread.run(HandlerThread.java:61)
```

JakeWharton:
We only call this method with the values 1-4 which are legal according to the Javadoc so I'm inclined to lean towards this being an Android bug.

narayank:
I don't think this is an android bug. The issue is that the latest version of jsr-166 sources (which ship with OpenJdk8 and 9) have a check that the core pool size isn't greater than the max pool size.  You should perhaps update the max pool size before you update the core pool size. In other words, if you're *increasing* the thread count you should do

setMaxPoolSize(N);
setCorePoolSize(N);

instead of

setCorePoolSize(N);
setMaxPoolSize(N);

I have applied a platform workaround (https://android-review.googlesource.com/#/c/208931/) but it might be reverted in a future release.

JakeWharton:
Got it. Will fix!

PR #1328, opened 2016-03-23
Set max pool size and core size according to Android SDK version.

cartierjf:
Fix in #1323 introduces a new bug related to #1319. There's a check in `ThreadPoolExecutor` that makes the new `setCorePoolSize(N)` `setMaxPoolSize(N)` order crash on Android < N:
```java
public void setMaximumPoolSize(int maximumPoolSize) {
    if (maximumPoolSize <= 0 || maximumPoolSize < corePoolSize)
        throw new IllegalArgumentException();
    this.maximumPoolSize = maximumPoolSize;
    if (workerCountOf(ctl.get()) > maximumPoolSize)
        interruptIdleWorkers();
    }
```
It causes an exception:
```
FATAL EXCEPTION: Picasso-Dispatcher
    Process: com.directed.ubi, PID: 12406
    java.lang.IllegalArgumentException
    at java.util.concurrent.ThreadPoolExecutor.setMaximumPoolSize(ThreadPoolExecutor.java:1633)
    at com.squareup.picasso.PicassoExecutorService.setThreadCount(PicassoExecutorService.java:81)
    at com.squareup.picasso.PicassoExecutorService.adjustThreadCount(PicassoExecutorService.java:44)
    at com.squareup.picasso.Dispatcher.performNetworkStateChange(Dispatcher.java:401)
    at com.squareup.picasso.Dispatcher$DispatcherHandler.handleMessage(Dispatcher.java:521)
    at android.os.Handler.dispatchMessage(Handler.java:102)
    at android.os.Looper.loop(Looper.java:148)
    at android.os.HandlerThread.run(HandlerThread.java:61)
```
To get it thrown, get the phone offline while loading an image.
I added a check for the Android version. Not sure it's the best way to do it, so feel free to suggest better solutions.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
@@ -17,6 +17,7 @@ package com.squareup.picasso;
 
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
+import android.os.Build;
 import android.telephony.TelephonyManager;
 
 import java.util.concurrent.Future;
--- a/picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
@@ -78,7 +78,7 @@ class PicassoExecutorService extends ThreadPoolExecutor {
   }
 
   private void setThreadCount(int threadCount) {
-    if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M) {
+    if (Build.VERSION.SDK_INT > 23) {
       setMaximumPoolSize(threadCount);
       setCorePoolSize(threadCount);
     } else {
--- a/picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
+++ b/picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
@@ -78,8 +78,13 @@ class PicassoExecutorService extends ThreadPoolExecutor {
   }
 
   private void setThreadCount(int threadCount) {
-    setMaximumPoolSize(threadCount);
-    setCorePoolSize(threadCount);
+    if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M) {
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
```diff
@@ -78,8 +78,13 @@ void adjustThreadCount(NetworkInfo info) {
   }
 
   private void setThreadCount(int threadCount) {
-    setMaximumPoolSize(threadCount);
-    setCorePoolSize(threadCount);
+    if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M) {
```
JakeWharton: You need to use "23" here, not the constant.

cartierjf:
Hi @JakeWharton, I've done the change. Might I ask why you prefer the number instead of the constant? Is it per convention?

JakeWharton:
It's because the project builds with Maven against an older version of Android API where the `M` constant isn't available.

picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
```diff
@@ -78,8 +78,13 @@ void adjustThreadCount(NetworkInfo info) {
   }
 
   private void setThreadCount(int threadCount) {
-    setMaximumPoolSize(threadCount);
-    setCorePoolSize(threadCount);
+    if (Build.VERSION.SDK_INT > 23) {
```
JakeWharton: You're missing an import for this class. That's why the build is failing.

cartierjf:
Ah ok got it. thanks

picasso/src/main/java/com/squareup/picasso/PicassoExecutorService.java
cartierjf: dammit, I was trying to go too fast :P I'll fix and commit.

cartierjf:
Ok I've added the import and I ran `mvn clean verify` and everything passes, so it should be fine.

JakeWharton:
Squashed and merged manually. Thanks!

PR #1385, opened 2016-05-17
Perform null check for canceling requests

### LINKED ISSUE #1384: More consistent null input checks
Some methods accept null when it's a clear programming error to do so, and they fail silently rather than alert the developer that they're doing something wrong. Many of these methods correspond to other methods (e.g. inverse methods with the same input) that do perform null checks, so making the null checks consistent across all related methods would be ideal. There are other methods that fail on null input when their corresponding methods do just fine with a null input, often falling back to some reasonable default behavior (like showing a placeholder image).

- `Picasso.cancelRequest`: `ImageView`, `Target`, and `RemoteViews` variants. Corresponding methods `RequestCreator.into` all perform null checks, and providing null here is a programming error, so I believe null checks should throw on all these methods.
- `Picasso.cancelTag`, `pauseTag`, `resumeTag`: corresponding method `RequestCreator.tag` performs a null check, and a null tag doesn't make sense, so I believe null checks should throw on all these methods.
- `Picasso.invalidate`: `Uri`, `String`, and `File` variants. Corresponding methods `Picasso.load` all allow null, though the `String` variant fails on a blank/empty input. I'd argue that the `String` variant is commonly used with input coming from an external source like a backend server and therefore should be more permissive - null/empty/blank strings all seem reasonable, especially since there's such a clear and easy fallback to show the placeholder image. `Uri` and `File` objects are both created on the client side so it seems reasonable to reject null in those cases and allow those with invalid paths internally to proceed to the usual fallback placeholder/error behavior. Whatever is decided for `load` should be duplicated in the corresponding `invalidate` method. I believe that the `String` versions should allow null/empty/blank, while the `Uri` and `File` versions should throw on null.

dnkoutso:
closes #1384 ?

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -203,11 +203,19 @@ public class Picasso {
 
   /** Cancel any existing requests for the specified target {@link ImageView}. */
   public void cancelRequest(ImageView view) {
+    // checkMain() is called from cancelExistingRequest()
+    if (view == null) {
+      throw new IllegalArgumentException("view cannot be null.");
+    }
     cancelExistingRequest(view);
   }
 
   /** Cancel any existing requests for the specified {@link Target} instance. */
   public void cancelRequest(Target target) {
+    // checkMain() is called from cancelExistingRequest()
+    if (target == null) {
+      throw new IllegalArgumentException("target cannot be null.");
+    }
     cancelExistingRequest(target);
   }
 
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -203,11 +203,19 @@
 
   /** Cancel any existing requests for the specified target {@link ImageView}. */
   public void cancelRequest(ImageView view) {
+    // checkMain() is called from cancelExistingRequest()
+    if (view == null) {
+      throw new AssertionError("view cannot be null.");
```
JakeWharton: I'm pretty sure we use IllegalArgumentException for these everywhere else.

JakeWharton:
lgtm otherwise

picasso/src/main/java/com/squareup/picasso/Picasso.java
dnkoutso: Will change.

dnkoutso:
Fixed.

JakeWharton:
:+1:

PR #1389, opened 2016-05-20
Remove unused parameter 'int networkPolicy' in RequestHandlers

NohSeho:


JakeWharton:
This is public API and cannot be changed without a major version release.

PR #1408, opened 2016-06-08
Do not clear LruCache with too large Bitmaps

NightlyNexus:


picasso/src/main/java/com/squareup/picasso/LruCache.java
```diff
@@ -73,8 +73,12 @@ public LruCache(int maxSize) {
 
     Bitmap previous;
     synchronized (this) {
+      int addedSize = Utils.getBitmapBytes(bitmap);
+      if (addedSize > maxSize) {
+        return;
```
JakeWharton: Outside synchronized? No reason to take a lock to return

NightlyNexus: Oh, yep. 👍  :p

JakeWharton:
Thanks!

PR #1409, opened 2016-06-09
Switch to gradle and AAR

hzsweers:
As discussed, this purposefully tries to limit the scope to just necessary changes. The android gradle plugin and build tools are updated, but the rest of the dependencies use the same as before. This configuration matches the ones used in sqldelight/rxbinding. The one thing I wasn't sure abut was emma.


picasso-pollexor/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+  compileSdkVersion rootProject.ext.compileSdkVersion
+  buildToolsVersion rootProject.ext.buildToolsVersion
+
+  compileOptions {
+    sourceCompatibility rootProject.ext.sourceCompatibilityVersion
+    targetCompatibility rootProject.ext.targetCompatibilityVersion
+  }
+
+  lintOptions {
+    textOutput 'stdout'
```
JakeWharton: I think this needs `textReport true` as well

JakeWharton: I think this needs `textReport true` as well

picasso-sample/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+  buildToolsVersion rootProject.ext.buildToolsVersion
+
+  compileOptions {
+    sourceCompatibility rootProject.ext.sourceCompatibilityVersion
+    targetCompatibility rootProject.ext.targetCompatibilityVersion
+  }
+
+  lintOptions {
+    lintConfig file('lint.xml')
+    textOutput 'stdout'
```
JakeWharton: `textReport true`

picasso/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+  compileSdkVersion rootProject.ext.compileSdkVersion
+  buildToolsVersion rootProject.ext.buildToolsVersion
+
+  compileOptions {
+    sourceCompatibility rootProject.ext.sourceCompatibilityVersion
+    targetCompatibility rootProject.ext.targetCompatibilityVersion
+  }
+
+  lintOptions {
+    textOutput 'stdout'
```
JakeWharton: `textReport true`

JakeWharton:
YOLO

PR #1412, opened 2016-06-09
Switch to support annotations on public APIs where appropriate

### LINKED ISSUE #1384: More consistent null input checks

hzsweers:
Also should resolve #1384 and removes the intellijAnnotations

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Downloader.java
+++ b/picasso/src/main/java/com/squareup/picasso/Downloader.java
@@ -17,6 +17,9 @@ package com.squareup.picasso;
 
 import android.graphics.Bitmap;
 import android.net.Uri;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
 import java.io.IOException;
 import java.io.InputStream;
 
@@ -32,7 +35,7 @@ public interface Downloader {
    * loading the bitmap.
    * @throws IOException if the requested URL cannot successfully be loaded.
    */
-  Response load(Uri uri, int networkPolicy) throws IOException;
+  @Nullable Response load(@NonNull Uri uri, int networkPolicy) throws IOException;
 
   /**
    * Allows to perform a clean up for this {@link Downloader} including closing the disk cache and
@@ -67,7 +70,7 @@ public interface Downloader {
      * @deprecated Use {@link RequestHandler} for directly loading {@link Bitmap} instances.
      */
     @Deprecated
-    public Response(Bitmap bitmap, boolean loadedFromCache) {
+    public Response(@NonNull Bitmap bitmap, boolean loadedFromCache) {
       if (bitmap == null) {
         throw new IllegalArgumentException("Bitmap may not be null.");
       }
@@ -85,7 +88,7 @@ public interface Downloader {
      * @deprecated Use {@link Response#Response(java.io.InputStream, boolean, long)} instead.
      */
     @Deprecated @SuppressWarnings("UnusedDeclaration")
-    public Response(InputStream stream, boolean loadedFromCache) {
+    public Response(@NonNull InputStream stream, boolean loadedFromCache) {
       this(stream, loadedFromCache, -1);
     }
 
@@ -100,7 +103,7 @@ public interface Downloader {
      * boolean)}.
      */
     @Deprecated @SuppressWarnings("UnusedDeclaration")
-    public Response(Bitmap bitmap, boolean loadedFromCache, long contentLength) {
+    public Response(@NonNull Bitmap bitmap, boolean loadedFromCache, long contentLength) {
       this(bitmap, loadedFromCache);
     }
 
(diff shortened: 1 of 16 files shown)


picasso/src/main/java/com/squareup/picasso/Downloader.java
```diff
@@ -32,7 +35,7 @@
    * loading the bitmap.
    * @throws IOException if the requested URL cannot successfully be loaded.
    */
-  Response load(Uri uri, int networkPolicy) throws IOException;
+  @Nullable Response load(@NonNull Uri uri, int networkPolicy) throws IOException;
```
JakeWharton: We allow `null`? I thought it was `Response` or throw only...

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -34,6 +39,7 @@
 import java.util.WeakHashMap;
 import java.util.concurrent.ExecutorService;
 
+import static android.R.attr.tag;
```
JakeWharton: Delete

picasso/src/main/java/com/squareup/picasso/Downloader.java
hzsweers: ¯\_(ツ)_/¯ I was just going by what the doc on line 34 said

picasso/src/main/java/com/squareup/picasso/Picasso.java
hzsweers: :O intelliJ 16's new auto-import-static

picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -411,13 +416,16 @@ public Builder purgeable() {
     }
 
     /** Decode the image using the specified config. */
-    public Builder config(Bitmap.Config config) {
+    public Builder config(@NonNull Bitmap.Config config) {
+      if (config == null) {
+        throw new IllegalArgumentException("tag == null");
```
JakeWharton: wrong message

picasso/src/main/java/com/squareup/picasso/Picasso.java
hzsweers: ce01e2d

picasso/src/main/java/com/squareup/picasso/Request.java
hzsweers: c88c020

picasso/src/main/java/com/squareup/picasso/Downloader.java
JakeWharton: Hmm... maybe returning null doesn't retry?

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -365,7 +376,7 @@ public RequestCreator load(int resourceId) {
    * @see #invalidate(String)
    * @see #invalidate(File)
    */
-  public void invalidate(Uri uri) {
+  public void invalidate(@NonNull Uri uri) {
```
JakeWharton: This should be nullable since `into(Uri)` is nullable.

JakeWharton: And obviously the check removed

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -379,9 +390,9 @@ public void invalidate(Uri uri) {
    * @see #invalidate(Uri)
    * @see #invalidate(File)
    */
-  public void invalidate(String path) {
+  public void invalidate(@Nullable String path) {
     if (path == null) {
-      throw new IllegalArgumentException("path == null");
+      return;
```
JakeWharton: Flip the `if` and move the invalidate call up

hzsweers: c3d2b2a

picasso/src/main/java/com/squareup/picasso/Picasso.java
hzsweers: 0106bef

PR #1414, opened 2016-06-09
Don't close cache if okhttp client instance is shared

### LINKED ISSUE #1346: Picasso Closes Shared OKHttp Cache on Shutdown
If a Picasso instance is created using a shared `OkHttpClient` it will close its `Cache` on `Picasso.shutdown`, which causes anything else using that `OkHttpClient` and its `Cache` to have problems. [PR that introduced the unwanted behavior](https://github.com/square/picasso/pull/713).

hzsweers:
Resolves #1346

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/OkHttp3Downloader.java
+++ b/picasso/src/main/java/com/squareup/picasso/OkHttp3Downloader.java
@@ -18,6 +18,7 @@ package com.squareup.picasso;
 import android.content.Context;
 import android.net.Uri;
 import android.support.annotation.NonNull;
+import android.support.annotation.VisibleForTesting;
 
 import java.io.File;
 import java.io.IOException;
@@ -32,6 +33,7 @@ import okhttp3.ResponseBody;
 public final class OkHttp3Downloader implements Downloader {
   private final Call.Factory client;
   private final Cache cache;
+  private boolean sharedClient = true;
 
   /**
    * Create new downloader that uses OkHttp. This will install an image cache into your application
@@ -70,6 +72,7 @@ public final class OkHttp3Downloader implements Downloader {
    */
   public OkHttp3Downloader(final File cacheDir, final long maxSize) {
     this(new OkHttpClient.Builder().cache(new Cache(cacheDir, maxSize)).build());
+    sharedClient = false;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


picasso/src/test/java/com/squareup/picasso/OkHttp3DownloaderTest.java
```diff
@@ -119,11 +120,17 @@
     }
   }
 
-  @Test public void shutdownClosesCache() throws Exception {
-    OkHttpClient client = new OkHttpClient();
-    Cache cache = new Cache(temporaryFolder.getRoot(), 100);
-    client.setCache(cache);
-    new OkHttpDownloader(client).shutdown();
+  @Test public void shutdownClosesCacheIfNotShared() throws Exception {
```
hzsweers: These tests were actually using OkHttpDownloader rather than OkHttp3Downloader before

picasso/src/main/java/com/squareup/picasso/OkHttp3Downloader.java
```diff
@@ -32,6 +33,7 @@
 public final class OkHttp3Downloader implements Downloader {
   private final Call.Factory client;
   private final Cache cache;
+  private boolean sharedClient = true;
```
JakeWharton: Make `final`?

hzsweers: I'm not sure how to do that actually in a fall through constructor like this. The current structure is that the second-to-last constructor knows it's making a default okhttpclient instance. The final constructor would need some way of knowing if the client it's given was created internally or not

picasso/src/test/java/com/squareup/picasso/OkHttp3DownloaderTest.java
```diff
@@ -30,6 +25,12 @@
 import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
 
+import okhttp3.OkHttpClient;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+import okio.Okio;
```
JakeWharton: We have a 'SquareAndroid' style [here](https://github.com/square/java-code-styles) if you want to stop screwin' with our order 😝

hzsweers: ooooo that's really useful. I'll use it in the future :)

PR #1417, opened 2016-06-10
Use getAllocationByteCount() for API 19+

### LINKED ISSUE #1072: Reconfigured Bitmaps will cause inconsistent reporting in LruCache
On >= 19, Bitmaps can be reconfigured, and getByteCount will return the configured byte count.  The following code will crash the LruCache:

        Cache cache = new LruCache(context);
        String key = "key";
        Bitmap original = Bitmap.createBitmap(5, 5, Bitmap.Config.ALPHA_8);
        cache.set(key, original);
        original.reconfigure(1, 1, original.getConfig());
        cache.clear(); // kaboom

This can be easily fixed in Utils.getBitmapBytes by adding getAllocationByteCount for >= 19, but Maven doesn't support building against 19, so we are waiting for https://github.com/square/picasso/pull/876

I'll add a test and a fix when ready; this is just a reminder.

hzsweers:
Should resolve #1072, but I don't know how to add a test for it. Robolectric barfs on it in both 2.0 and 3.1

CC @NightlyNexus


dnkoutso:
👍

JakeWharton:
Not a lot of risk in leaving this untested so it's fine.

PR #1423, opened 2016-06-18
API 24

### LINKED ISSUE #903: Annotate resize(int, int) with @Px
I'm aware this is more of a useful feature rather than bug. There is resizeDimen(int,int) but when I made a mistake and called resize(R.dimen,R.dimen) Picasso fails to load image. Is there an option to supply @DimenRes to arguments or any other way of checking in Runtime that input is too big and then throwing IllegalArgumentException?

hzsweers:
> Seems like the timetable for v24 would be when N is released, which might be months away still

- A pessimistic fool.

Tested with Travis and worked fine as well (!!): https://github.com/hzsweers/picasso/pull/1

Also resolves #903

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ android:
   components:
     - tools
     - platform-tools
-    - build-tools-23.0.3
+    - build-tools-24.0.0
     - extra-android-m2repository
-    - android-23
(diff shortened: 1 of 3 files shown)


dnkoutso:
👍

PR #1430, opened 2016-06-26
Specify default timeouts for okhttp3.OkHttpClient

kaorimatz:
This relates to #102.


JakeWharton:
OkHttp 3 has default timeouts so they do not need set.

kaorimatz:
I see, thanks!

PR #1431, opened 2016-06-26
Avoid unnecessary call to trimToSize when invalidating cache

kaorimatz:
Invalidating cached bitmaps doesn't increase the cache size, so I believe we don't need to call trimToSize.


JakeWharton:
Nice!

PR #1433, opened 2016-06-27
Fix wrong minSdkVersion in Sample

ncornette:
minSdkVersion was defined twice, second definition was in fact the targetSdkVersion
 
With minSdkVersion restored, Lint reported NewApi issue in styles.xml which should have been ignored but path was incorrect. Travis is now happy!


JakeWharton:
Thanks

Issue #1446, opened 2016-07-09
fit() does not defer if the ImageView already has a height and width


suau:
I'm using Picasso in a RecyclerView with a custom ImageView (ArImageView).
ArImageView allows setting a aspect ratio for the view (calls `requestLayout()`), which together one side of the ImageView will determine the height/width of the final view in `onMeasure`.
Picassos `fit()` will only defer loading when either the height or the width is 0, this does not work well with re-used and re-measured Views (like ArImageView).
This is not specific to my custom view and will occur with any re-used ImageView.

NightlyNexus:
Re-open this until a fix is found?

PR #1447, opened 2016-07-09
Fix fit() deferred loading for re-sized views

suau:
Fix issue: https://github.com/square/picasso/issues/1446

Picassos fit() will only defer loading when either the height or the width is 0, this does not work well with re-used and re-measured Views.
This fix additionally checks if a layout has been requested for the given ImageView when deferring and also will not start loading until layout has been done.


dnkoutso:
Can you please add a test in `DeferredRequestCreatorTest`?

JakeWharton:
Thanks!

suau:
I have to thank for this great library.

JakeWharton:
I'm really trying to find time to finish the current release. It will
be.... this year for sure :D

swankjesse:
Bad news: this breaks `fit()`.
https://github.com/square/picasso/issues/1749

I’m reverting this in #1763. I hope to get it working with fit() and then I’ll put it back.

PR #1456, opened 2016-07-19
Fixes onlyScaleDown() when targetWidth or targetHeight is 0

### LINKED ISSUE #1432: onlyScaleDown() does not work if using "magic" 0 value in resize()
in `BitmapHunter.java`

```
  private static boolean shouldResize(boolean onlyScaleDown, int inWidth, int inHeight,
      int targetWidth, int targetHeight) {
    return !onlyScaleDown || inWidth > targetWidth || inHeight > targetHeight;
  }
```
obviously if I use 0 for width or height then the image is always resized since targetWidth or targetHeight is 0

mblcdr:
Resolves #1432

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
+++ b/picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
@@ -873,6 +873,54 @@ public class BitmapHunterTest {
     assertThat(shadowBitmap.getCreatedFromBitmap()).isNotSameAs(source);
   }
 
+  @Test public void onlyScaleDownOriginalSmallerWidthIs0() {
+    Bitmap source = Bitmap.createBitmap(50, 50, ARGB_8888);
+    Request data = new Request.Builder(URI_1).resize(0, 60).onlyScaleDown().build();
+    Bitmap result = transformResult(data, source, 0);
+    assertThat(result).isSameAs(source);
+
+    ShadowBitmap shadowBitmap = shadowOf(result);
+    assertThat(shadowBitmap.getCreatedFromBitmap()).isNull();
+  }
+
+  @Test public void onlyScaleDownOriginalSmallerHeightIs0() {
+    Bitmap source = Bitmap.createBitmap(50, 50, ARGB_8888);
+    Request data = new Request.Builder(URI_1).resize(60, 0).onlyScaleDown().build();
+    Bitmap result = transformResult(data, source, 0);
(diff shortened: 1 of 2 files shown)


picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -873,6 +873,28 @@
...
 
+  @Test public void onlyScaleDownResizeWidthIs0() {
+    Bitmap source = Bitmap.createBitmap(50, 50, ARGB_8888);
+    Request data = new Request.Builder(URI_1).resize(0, 60).onlyScaleDown().build();
+    Bitmap result = transformResult(data, source, 0);
+    assertThat(result).isSameAs(source);
+
+    ShadowBitmap shadowBitmap = shadowOf(result);
+    assertThat(shadowBitmap.getCreatedFromBitmap()).isNull();
+    assertThat(shadowBitmap.getCreatedFromBitmap()).isNotSameAs(source);
```
JakeWharton: This test logic is wrong (it looks copy/pasted). If we already check the value is null there's no reason to check it against something else.

I suspect it was meant to be `assertThat(result).isNotSameAs(source)` in the original test and here it should be `isSameAs`.

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -873,6 +873,28 @@
...
+  }
+
+  @Test public void onlyScaleDownResizeHeightIs0() {
+    Bitmap source = Bitmap.createBitmap(50, 50, ARGB_8888);
+    Request data = new Request.Builder(URI_1).resize(60, 0).onlyScaleDown().build();
```
JakeWharton: Can you also add two tests where it actually still needs scaled? Like for `40, 0` and `0, 40`?

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
mblcdr: I took the code from `onlyScaleDownOriginalSmaller()` which has the same logic. Is that test faulty, too?

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
mblcdr: Will do.

JakeWharton:
Thanks!

PR #1269, opened 2016-01-18
Translate MediaStore.Image.ORIENTATION to ExifInterface.ORIENTATION to keep orientation processing proper.

RobertM-BLS:
MediaStoreRequestHandler: read MediaStore.Image.ORIENTATION value and translate it to ExifInterface.ORIENTATION_* value.

Picasso determines orientation according to ExifInterface.ORIENTATION_*. All possible values in MediaStore.Image.ORIENTATION are: 0, 90, 180, 270 (according to: https://developer.android.com/reference/android/provider/MediaStore.Images.ImageColumns.html#ORIENTATION). The values differ, thus orientation read from MediaStore must be translated to ExifInterface.ORIENTATION_* value.

Updated RequestHandler::getExifOrientation description.

PR #1459, opened 2016-07-20
Fixed EXIF orientation in MediaStoreRequestHandler

st1hy:
Merged changes from #1269 into this and fixed merge conflicts.

In my test even those changes do not fix orientation problems 100%, because at least on my phones orientation from content resolver do not always match actual EXIF in file. Therefore I've added reading EXIF directly from file. This data tends to be correct pretty much always. 

I've left reading from content resolver as a fall-back if there is an exception when reading file.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/MediaStoreRequestHandler.java
+++ b/picasso/src/main/java/com/squareup/picasso/MediaStoreRequestHandler.java
@@ -23,15 +23,16 @@ import android.graphics.BitmapFactory;
 import android.media.ExifInterface;
 import android.net.Uri;
 import android.provider.MediaStore;
+
 import java.io.IOException;
 
 import static android.content.ContentResolver.SCHEME_CONTENT;
 import static android.content.ContentUris.parseId;
 import static android.provider.MediaStore.Images;
-import static android.provider.MediaStore.Video;
 import static android.provider.MediaStore.Images.Thumbnails.FULL_SCREEN_KIND;
 import static android.provider.MediaStore.Images.Thumbnails.MICRO_KIND;
 import static android.provider.MediaStore.Images.Thumbnails.MINI_KIND;
+import static android.provider.MediaStore.Video;
 import static com.squareup.picasso.MediaStoreRequestHandler.PicassoKind.FULL;
 import static com.squareup.picasso.MediaStoreRequestHandler.PicassoKind.MICRO;
 import static com.squareup.picasso.MediaStoreRequestHandler.PicassoKind.MINI;
@@ -40,6 +41,9 @@ import static com.squareup.picasso.Picasso.LoadedFrom.DISK;
 class MediaStoreRequestHandler extends ContentStreamRequestHandler {
   private static final String[] CONTENT_ORIENTATION = new String[] {
       Images.ImageColumns.ORIENTATION
+  };
+  private static final String[] CONTENT_DATA = new String[] {
+          Images.ImageColumns.DATA
   };
 
   MediaStoreRequestHandler(Context context) {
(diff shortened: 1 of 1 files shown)


cjwilliams24680:
Is there any update on this?

JakeWharton:
Updates are posted as comments to issues so if there's no new comments then
there's no new updates.

lmller:
this fix doesn't seem to work, since `BitmapHunter.transformResult(Request data, Bitmap result, int exifRotation)` does a ` matrix.preRotate(exifRotation);` so my image is now rotated by 8 degrees, instead of `ORIENTATION_ROTATE_270` (which is 8)

picasso/src/main/java/com/squareup/picasso/MediaStoreRequestHandler.java
```diff
@@ -103,6 +107,37 @@ static PicassoKind getPicassoKind(int targetWidth, int targetHeight) {
...
+      cursor = contentResolver.query(uri, CONTENT_DATA, null, null, null);
+      if (cursor == null || !cursor.moveToFirst()) {
+        return 0;
+      }
+      String filePath = cursor.getString(0);
+      ExifInterface exifInterface = new ExifInterface(filePath);
+      return exifInterface.getAttributeInt(ExifInterface.TAG_ORIENTATION,
+              ExifInterface.ORIENTATION_UNDEFINED);
+
+    } catch (Exception ignored) {
```
RyPope: Instead of catching and ignoring all exceptions, I think the proper way would be to retrieve the mime type of the image and only attempt to get orientation from exif metadata if the type supports exif (jpeg), otherwise return undefined.

picasso/src/main/java/com/squareup/picasso/MediaStoreRequestHandler.java
```diff
@@ -103,6 +107,37 @@ static PicassoKind getPicassoKind(int targetWidth, int targetHeight) {
...
 
   static int getExifOrientation(ContentResolver contentResolver, Uri uri) {
+    int exifOrientation = getExitOrientationFromFile(contentResolver, uri);
+    if (exifOrientation == ExifInterface.ORIENTATION_UNDEFINED) {
+      exifOrientation = getExifOrientationFromContentResolver(contentResolver, uri);
+    }
+    return exifOrientation;
+  }
+
+  static int getExitOrientationFromFile(ContentResolver contentResolver, Uri uri) {
```
RyPope: exif*. I think it would be more clear to parse the file path and pass it in here instead of a uri. It seems incorrect to call this "getExifOrientationFromFile" and not pass a file.

konstantin-n:
> this fix doesn't seem to work, since `BitmapHunter.transformResult(Request data, Bitmap result, int exifRotation)` does a ` matrix.preRotate(exifRotation);`
> so my image is now rotated by 8 degrees, instead of `ORIENTATION_ROTATE_270` (which is 8)

May be to late, but `exifRotation` takes value from `getExifRotation(int orientation)`:
```
static int getExifRotation(int orientation) {
    int rotation;
    switch (orientation) {
      case ORIENTATION_ROTATE_90:
      case ORIENTATION_TRANSPOSE:
        rotation = 90;
        break;
      case ORIENTATION_ROTATE_180:
      case ORIENTATION_FLIP_VERTICAL:
        rotation = 180;
        break;
      case ORIENTATION_ROTATE_270:
      case ORIENTATION_TRANSVERSE:
        rotation = 270;
        break;
      default:
        rotation = 0;
    }
    return rotation;`
  }
```
Those it can be only 0, 90, 180, 270.
Details - #1269
On my opinion the fix from #1269 sufficient. Supplying `content` as Uri for image we rely on data from `ImageProvider`(on this case we don't care about **exif** info). When supplying `file` as uri we rely on **exif** info. There is still issue on library - Images received from MediaProvider.Images ignore rotation after decoding(`getExifRotation(int orientation)` always returns 0).

PR #1238, opened 2015-12-21
centerCrop() align gravity options

jskierbi:
I've added overloaded centerCrop() method with align gravity parameter (```android.view.Gravity```).

When gravity parameter is supplied, image is aligned first, then cropped.

Example call:
```
Picasso.with(context)
    .load(url)
    .fit()
    .centerCrop(Gravity.TOP)
    .into(imageView);
```

Parameterless ```centerCrop()``` is synonymous to ```centerCrop(Gravity.CENTER)```

Example results:

* ```centerCrop(Gravity.TOP | Gravity.START)```
top_start

* ```centerCrop(Gravity.BOTTOM | Gravity.END)```
bottom_end

ImageViews used in examples:
```
<ImageView
    android:id="@+id/image_horizontal"
    android:layout_width="200dp"
    android:layout_height="100dp" 
    />
    
<ImageView
    android:id="@+id/image_vertial"
    android:layout_width="100dp"
    android:layout_height="200dp" 
    />
```

PR #1469, opened 2016-08-09
Add align gravity options to centerCrop() method

JakeWharton:
Rebased and added tests. Closes #1238.


JakeWharton:
Choke yourself checkstyle

PR #1509, opened 2016-10-19
Fixed: Sample App crashes on devices with Android 6.0 and higher.

NahomAgidew:
Fixed: Sample App crashes on devices with Android 6.0 and higher due to runtime permissions. I added an additional step to ask user for permission to read contacts photos at runtime.


jrodbx:
Fixed in #1767

PR #1516, opened 2016-10-26
Add a flag to always transform remote images with thumbor when set.

runningcode:


JakeWharton:
Seems fine. Test?

runningcode:
Added tests!

PR #1522, opened 2016-11-01
Add Callback parameter for loading into RemoteViews

NightlyNexus:
I have use cases, but this is another two options for the most important part of the API.
Thanks for considering it!


picasso/src/main/java/com/squareup/picasso/RemoteViewsAction.java
```diff
@@ -27,25 +27,41 @@
 abstract class RemoteViewsAction extends Action<RemoteViewsAction.RemoteViewsTarget> {
   final RemoteViews remoteViews;
   final int viewId;
+  Callback callback;
```
NightlyNexus: I couldn't tell if `@Nullable` was appropriate to start adding, so I omitted it.

JakeWharton: Both would have been fine. We only got support annotations recently.

PR #1546, opened 2016-12-01
Request permissions at runtime to fix contact and gallery samples

jaredrummler:


jrodbx:
Fixed in https://github.com/square/picasso/pull/1767

PR #1552, opened 2016-12-12
Add supported package name or apk path loading app icon

fushenghua:
Add supported package name or apk path loading app icon,need this feature.
Thanks for considering it!


JakeWharton:
The whole point of the `RequestHandler` abstraction is so that you can install new handlers like this without them being part of the core library. At this time we don't want to be adding new ones and I'm not convinced this is a general enough use case to warrant inclusion in the core. If you want you can publish this as a separate library for others to include.

PR #1569, opened 2017-01-05
Pass failure exception to callback and target.

### LINKED ISSUE #617: Getting the exact error message in Picasso Callback onError
Hi,

I have an app in production which uses Picasso to load a variety of image URLs. 
Some of these URLs point to "invalid" images which cannot be rendered by Picasso. So far, I have handled these exceptions by indicating what the error was to the end-user.

My dilemma lies in that these "cannot render image" exceptions are all generic. Neverminding the problems I can predict in advance (i.e. check if the image URL has a valid MIME type) my app can't articulate the precise reason why an image failed to load. To date, this lack of specificity has confused my users over whether the problem is with my app or the server from which the images are being loaded.

Are there plans to percolate the rendering exception that Picasso encounters to the callback for Picasso.load.into?

Looking forward to hearing other's thoughts on this matter.

### LINKED ISSUE #1548: Picasso doesn't write exceptions to stacktrace
I stumbled upon quite of a minor issue in Picasso. I am loading images from storage to Target using Picasso and I didn't have READ_EXTERNAL_STORAGE permission in my manifest so I got a SecurityException caught in BitmapHunter (I know that it should be obvious to turn on the permission but I just forgot about it). I had to debug quite a lot to get into it because even with logging enabled Picasso didn't show me the stacktrace. I strongly suggest to change it in the upcoming versions.

That kind of logging doesn't tell me much about the problem:

    D/Picasso: Main        created      [R0] Request{content://media/external/images/media/1623}
    D/Picasso: Dispatcher  enqueued     [R0]+7ms 
    D/Picasso: Hunter      executing    [R0]+7ms 
    D/Picasso: Dispatcher  batched      [R0]+42ms for error
    D/Picasso: Dispatcher  delivered    [R0]+244ms 
    D/Picasso: Main        errored      [R0]+244ms

JakeWharton:
Closes #1548. Closes #617.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Action.java
+++ b/picasso/src/main/java/com/squareup/picasso/Action.java
@@ -63,7 +63,7 @@ abstract class Action<T> {
 
   abstract void complete(Bitmap result, Picasso.LoadedFrom from);
 
-  abstract void error();
+  abstract void error(Exception e);
 
   void cancel() {
     cancelled = true;
--- a/picasso/src/main/java/com/squareup/picasso/Callback.java
+++ b/picasso/src/main/java/com/squareup/picasso/Callback.java
@@ -18,14 +18,14 @@ package com.squareup.picasso;
 public interface Callback {
   void onSuccess();
 
-  void onError();
+  void onError(Exception e);
 
   class EmptyCallback implements Callback {
 
     @Override public void onSuccess() {
     }
 
-    @Override public void onError() {
+    @Override public void onError(Exception e) {
     }
   }
 }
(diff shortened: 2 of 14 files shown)


dnkoutso:
🚢

dann1609:
Hey I'm updating library but callback is not updating. Right now i have version 2.5.2 someone have the same problem?

NightlyNexus:
@dann1609 This change is unreleased. You can use the 3.0.0 snapshots if you need it.
```gradle
repositories {
  maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
}
dependencies {
  compile 'com.squareup.picasso:picasso:3.0.0-SNAPSHOT'
}
```

PR #1591, opened 2017-02-24
Plumb Okio stream types into request handlers.

JakeWharton:
This simplifies a ton of byte-management operations like checking for webp or even converting to a byte[].

@swankjesse

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
@@ -29,6 +29,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
+import okio.BufferedSource;
+import okio.Okio;
+import okio.Source;
 
 import static android.media.ExifInterface.ORIENTATION_FLIP_HORIZONTAL;
 import static android.media.ExifInterface.ORIENTATION_FLIP_VERTICAL;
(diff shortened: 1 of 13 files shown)


JakeWharton:
Oh no it's that stupid Robolectric bug...

picasso/src/main/java/com/squareup/picasso/BitmapHunter.java
```diff
@@ -119,37 +122,39 @@
...
       if (calculateSize) {
         BitmapFactory.decodeByteArray(bytes, 0, bytes.length, options);
         RequestHandler.calculateInSampleSize(request.targetWidth, request.targetHeight, options,
             request);
       }
       return BitmapFactory.decodeByteArray(bytes, 0, bytes.length, options);
     } else {
+      InputStream stream = bufferedSource.inputStream();
       if (calculateSize) {
+        // TODO use an InputStream that buffers with Okio...
```
swankjesse: yeah

picasso/src/main/java/com/squareup/picasso/Utils.java
```diff
@@ -283,25 +276,15 @@ static boolean hasPermission(Context context, String permission) {
...
-    byte[] fileHeaderBytes = new byte[WEBP_FILE_HEADER_SIZE];
-    boolean isWebPFile = false;
-    if (stream.read(fileHeaderBytes, 0, WEBP_FILE_HEADER_SIZE) == WEBP_FILE_HEADER_SIZE) {
-      // If a file's header starts with RIFF and end with WEBP, the file is a WebP file
-      isWebPFile = WEBP_FILE_HEADER_RIFF.equals(new String(fileHeaderBytes, 0, 4, "US-ASCII"))
-          && WEBP_FILE_HEADER_WEBP.equals(new String(fileHeaderBytes, 8, 4, "US-ASCII"));
+  static boolean isWebPFile(BufferedSource source) throws IOException {
+    if (source.request(WEBP_FILE_HEADER_SIZE)) {
+      Buffer buffer = source.buffer();
+      if (buffer.rangeEquals(0, WEBP_FILE_HEADER_RIFF)
```
swankjesse: you could potentially call `source.rangeEquals()` right? Then you don’t even need to call `request()` above

swankjesse: ... and that won’t request a 2nd byte if the first byte isn’t `R` or W`

JakeWharton: Ah, yeah. I was so worried about not consuming the byes I forgot about that.

JakeWharton: https://github.com/square/picasso/pull/1593

Issue #941, opened 2015-03-20
onlyScaleDown should be able to use with fit


Relaxuma:

dnkoutso:
Please provide a PR with a test to support the change. No plans for this now.

edwardaa:
+1
It's useful, please

mpost:
+1

rimeissner:
it should work to combine `resize(100, 100).onlyScaleDown().fit()` correct?

PR #1600, opened 2017-03-12
Allow onlyScaleDown() with fit() method

kwangsu61:
onlyScaleDown() method is so useful to avoid useless bitmap scale up for memory usage and performance.

This PR allows that onlyScaleDown() works with fit() method like centerCrop()/centerInside().

Refer to #941

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso/Request.java
@@ -390,9 +390,6 @@ public final class Request {
      * specified by {@link #resize(int, int)}.
      */
     public Builder onlyScaleDown() {
-      if (targetHeight == 0 && targetWidth == 0) {
-        throw new IllegalStateException("onlyScaleDown can not be applied without resize");
-      }
       onlyScaleDown = true;
       return this;
     }
@@ -499,6 +496,10 @@ public final class Request {
       if (centerInside && (targetWidth == 0 && targetHeight == 0)) {
         throw new IllegalStateException(
             "Center inside requires calling resize with positive width and height.");
+      }
+      if (onlyScaleDown && (targetWidth == 0 && targetHeight == 0)) {
+        throw new IllegalStateException(
+            "onlyScaleDown requires calling resize with positive width and height.");
       }
       if (priority == null) {
         priority = Priority.NORMAL;
--- a/picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
+++ b/picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
@@ -402,6 +402,16 @@ public class RequestCreatorTest {
   }
 
   @Test
+  public void intoImageViewWithFitAndOnlyScaleDown() {
+    ImageView target = mockFitImageViewTarget(true);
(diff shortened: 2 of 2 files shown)


magneticflux-:
It seems like an arbitrary restriction, I'd love this change! `.fit()` just uses `.resize()` behind the scenes, so I don't seen any problems using `.onlyScaleDown()` with it.

picasso/src/main/java/com/squareup/picasso/Request.java
```diff
@@ -500,6 +497,10 @@ public Request build() {
         throw new IllegalStateException(
             "Center inside requires calling resize with positive width and height.");
       }
+      if (onlyScaleDown && (targetWidth == 0 && targetHeight == 0)) {
```
magneticflux-: Should this be `if(deferred || (targetWidth == 0 && targetHeight == 0))`?

JakeWharton: And this check should be moved back to the `onlyScaleDown` method.

picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
```diff
@@ -555,6 +565,11 @@ public void intoTargetNoResizeWithCenterInsideOrCenterCropThrows() {
       fail("Center inside with unknown height should throw exception.");
     } catch (IllegalStateException ignored) {
     }
+    try {
+      new RequestCreator(picasso, URI_1, 0).onlyScaleDown().into(mockTarget());
+      fail("onlyScaleDown with unknown height should throw exception.");
```
magneticflux-: Shouldn't this fail when `.onlyScaleDown()` is called without a call to `.fit()` or `.resize(int, int)`?

picasso/src/test/java/com/squareup/picasso/RequestCreatorTest.java
```diff
@@ -544,7 +554,7 @@ public void intoRemoteViewsNotificationWithFitThrows() {
   }
 
   @Test
-  public void intoTargetNoResizeWithCenterInsideOrCenterCropThrows() {
+  public void intoTargetNoResizeWithCenterInsideOrCenterCropOrOnlyScaleDownThrows() {
```
magneticflux-: This should probably also be named `intoTargetNoResizeOrFitWithCenterInsideOrCenterCropOrOnlyScaleDownThrows`.

picasso/src/main/java/com/squareup/picasso/Request.java
magneticflux-: @JakeWharton This method is in the `Request` class, which seems to be in line with the other checks such as for null lists, invalid properties, and center[Crop|Inside] checks.

magneticflux-: Ahh, the `Request` class has no knowledge of fitting, which is done in the `RequestCreator` class. I'm going to make my own pull request that will hopefully fix these issues cleanly.

magneticflux-:
@JakeWharton I have an alternative pull request, #1623, that I believe fixes the issues with this one.

PR #1610, opened 2017-04-06
Adding documentation about using Animatable as a placeholder

teekirol:
I thought it would be helpful to include this in the documentation.


JakeWharton:
This would be nice to have on the `placeholder` and `error` methods. I don't think it needs put on the website because it's quite specific.

PR #1614, opened 2017-04-14
added a comment to explicit what drives the animation

martinbonnin:
I spent a couple of hours trying to figure out how this worked today so hopefully this will save some time for other people.


JakeWharton:
Thanks!

PR #1627, opened 2017-05-22
Use prepareToDraw to upload Bitmaps to GPU.

### LINKED ISSUE #1620: Use Bitmap#prepareToDraw to upload Bitmaps to GPU in advance
Hey, I work on UI Toolkit Graphics in Android. Wanted to pass on an optimization you can enable for displaying Bitmaps.
 
Android displays Bitmaps as OpenGL textures, and the first time a Bitmap is displayed in a frame, it’s uploaded to the GPU. That can take several milliseconds, but it’s necessary to display the image with the GPU.
 
In Android N, we added behavior to Bitmap#prepareToDraw() to send an async message to RenderThread to pre-upload. RenderThread uploads these to the GPU when it expects to be idle between frames, instead of while drawing the first frame that uses the Bitmap. If you can do this after decoding, you can make it much less likely that frames are dropped due to uploading.
 
In Systrace, you can see the problem by looking for sections labeled ‘Upload \<w\>x\<h\> Texture’. If those are happening in ‘DrawFrame’ on the RenderThread, inside ‘syncFrameState,’ they haven’t been pre-uploaded, and are on the critical path.
 
The drawback is that if you call it on a bitmap before you modify it, or if it doesn’t get drawn, you’ll be wasting the time spent uploading. If you’re reasonably sure the Bitmap in question will be displayed, it’s still a great way to prefetch the upload work on RenderThread.
 
Behavior added in Android N:
https://android.googlesource.com/platform/frameworks/base/+/4387190d8ec9fe4e953fcfeb093a644b82cf85ed
 
For reference, Glide implemented this here: https://github.com/bumptech/glide/commit/dce95506f138bd877aaed1183958f5e186a2bf

It's safe to call on older versions, so you don't need a version check. The method existed prior to N, but was a noop for normal Bitmaps.

NightlyNexus:
Closes #1620.

Did I miss anything?

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Dispatcher.java
+++ b/picasso/src/main/java/com/squareup/picasso/Dispatcher.java
@@ -431,6 +431,9 @@ class Dispatcher {
     if (hunter.isCancelled()) {
       return;
     }
+    if (hunter.result != null) {
+      hunter.result.prepareToDraw();
+    }
     batch.add(hunter);
     if (!handler.hasMessages(HUNTER_DELAY_NEXT_BATCH)) {
       handler.sendEmptyMessageDelayed(HUNTER_DELAY_NEXT_BATCH, BATCH_DELAY);


picasso/src/main/java/com/squareup/picasso/ImageViewAction.java
```diff
@@ -44,6 +44,8 @@
       return;
     }
 
+    result.prepareToDraw();
```
JakeWharton: This seems pretty late to call this. Can we do it before posting to the main thread?

NightlyNexus: Yeah.
What about https://github.com/square/picasso/blob/a834e259fba01971651a3cd684ea5370f3769de0/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java#L176
```java
if (action instanceof ImageViewAction || action instanceof RemoteViewsAction) {
  result.prepareToDraw();
}
```

dnkoutso: At least on line 176 we still don't know if we have a result. This should be done if we know a result was received e.g. on line https://github.com/square/picasso/blob/a834e259fba01971651a3cd684ea5370f3769de0/picasso/src/main/java/com/squareup/picasso/BitmapHunter.java#L180 instead.

Furthermore, we could minimize the window to only the point where we know the bitmap will have a high chance of actually being rendered. 

Perhaps right before we batch the results to send them to main thread here https://github.com/square/picasso/blob/master/picasso/src/main/java/com/squareup/picasso/Dispatcher.java#L430

NightlyNexus: Nice, thank you. I _think_ I did that now.
apologies for the extreme delay.

Issue #1634, opened 2017-06-30
Remove the Context from Picasso.with()


SalvatoreT:
Follow the [trick that Firebase uses](https://firebase.googleblog.com/2016/12/how-does-firebase-initialize-on-android.html).

Store a reference to this application context and then the first time you called `Picasso.with()`, either use the instance you already set or create one on-the-fly. We already do that inside `with(Context)`, but we use the `Context` you pass in instead of one that we steal from a content provider.

- [x]  copy that content provider hack to set the application context onto `Picasso`
- [x]  change `with` to remove the argument and use that application context we set

NightlyNexus:
Not everybody wants a ContentProvider or static instance of a Picasso in his or her app.

JakeWharton:
The instance is still lazy. You can tools: remove the provider if you want, though all it does is stash a reference so it doesn't add startup overhead or anything.

PR #1635, opened 2017-06-30
Retrieve Context from a ContentProvider

SalvatoreT:
Remove the need to have the context passed-in on `Picasso.with()`.

https://github.com/square/picasso/issues/1634

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -29,7 +29,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.widget.ImageView;
 import android.widget.RemoteViews;
-
 import java.io.File;
 import java.lang.ref.ReferenceQueue;
 import java.util.ArrayList;
@@ -59,7 +58,7 @@ import static com.squareup.picasso.Utils.log;
 /**
  * Image downloading, transformation, and caching manager.
  * <p>
- * Use {@link #with(android.content.Context)} for the global singleton instance or construct your
+ * Use {@link #with()} for the global singleton instance or construct your
  * own instance with {@link Builder}.
  */
 public class Picasso {
@@ -674,16 +673,9 @@ public class Picasso {
    * {@link Picasso} instance. You can either use this directly or by setting it as the global
    * instance with {@link #setSingletonInstance}.
    */
-  public static Picasso with(@NonNull Context context) {
-    if (context == null) {
-      throw new IllegalArgumentException("context == null");
-    }
+  public static Picasso with() {
     if (singleton == null) {
-      synchronized (Picasso.class) {
-        if (singleton == null) {
-          singleton = new Builder(context).build();
-        }
-      }
+      throw new IllegalStateException("Singleton was not initialized.");
     }
     return singleton;
   }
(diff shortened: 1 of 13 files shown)


SalvatoreT:
@JakeWharton, here's a first swing at it.

NightlyNexus:
Not everybody wants a ContentProvider or static instance of a Picasso in his or her app.

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -674,16 +673,9 @@ void shutdown() {
...
-  public static Picasso with(@NonNull Context context) {
-    if (context == null) {
-      throw new IllegalArgumentException("context == null");
-    }
+  public static Picasso with() {
     if (singleton == null) {
-      synchronized (Picasso.class) {
-        if (singleton == null) {
-          singleton = new Builder(context).build();
-        }
```
JakeWharton: We should still do the initialization here. Just stash the Context in a static field on this class from the content provider.

picasso/src/main/java/com/squareup/picasso/PicassoProvider.java
```diff
@@ -0,0 +1,55 @@
...
+package com.squareup.picasso;
+
+import android.content.ContentProvider;
+import android.content.ContentValues;
+import android.database.Cursor;
+import android.net.Uri;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+public class PicassoProvider extends ContentProvider {
```
JakeWharton: nit: final

JakeWharton: Also can you annotate this with `@RestrictTo(LIBRARY)` as an extra measure.

picasso/src/main/java/com/squareup/picasso/Picasso.java
JakeWharton: Actually stash the Context inside a static on the content provider itself. That will prevent this class from initializing until it's referenced.

picasso/src/main/java/com/squareup/picasso/PicassoProvider.java
```diff
@@ -0,0 +1,55 @@
...
+  @Override public boolean onCreate() {
+    Picasso.setSingletonInstance(new Picasso.Builder(getContext()).build());
```
NightlyNexus: if you want to override the singleton instance, then, you need to remove the ContentProvider from the manifest, right?
or is there another way?

SalvatoreT:
@JakeWharton & @NightlyNexus, ready for a second pass.

- [ ] squash commits

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -689,6 +688,16 @@ public static Picasso with(@NonNull Context context) {
   }
 
   /**
+   * The global default {@link Picasso} instance.
+   * <p>
+   * This instance is created by passing a global context from startup into {@link #with(Context)}.
+   * </p>
+   */
+  public static Picasso with() {
+    return with(PicassoProvider.context);
```
NightlyNexus: nice, still lazy and still lets setSingletonInstance work fine without disabling anything.

pyricau:
Suggestion: should the name be something else than `with()` ? "with context" makes sense, "with" by itself is a bit weird.

JakeWharton:
This was discussed in the follow-up where I removed the content version
completely.

PR #1636, opened 2017-07-01
Finish migration to argument-less with() method.

JakeWharton:


JakeWharton:
cc @SalvatoreT

SalvatoreT:
Does the name "`with`" still make sense now that there's no context directly passed-in?

JakeWharton:
No. Probably should be `get()`, but it'd also be great to figure out a way to just make `load()`, etc. static methods so you don't even need `get()`.

JakeWharton:
@dnkoutso can you review? weigh in?

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -673,31 +673,21 @@ void shutdown() {
...
-    if (context == null) {
-      throw new IllegalArgumentException("context == null");
-    }
+  public static Picasso with() {
     if (singleton == null) {
       synchronized (Picasso.class) {
         if (singleton == null) {
-          singleton = new Builder(context).build();
+          if (PicassoProvider.context == null) {
+            throw new IllegalStateException("context == null");
```
NightlyNexus: Drive-by thought:
This would only happen if you remove the ContentProvider in your manifest, right?
Should the error message suggest checking for that?

PR #1641, opened 2017-07-09
Fix permission bug for Android 6.0

CoXier:
Sample has not supported Android M yet.

### DIFF
--- a/picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
+++ b/picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
@@ -16,9 +16,13 @@
  */
 package com.example.picasso;
 
+import android.Manifest;
+import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
 import android.support.v4.content.Loader;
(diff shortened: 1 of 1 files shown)


picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
```diff
@@ -30,6 +34,9 @@
     implements LoaderManager.LoaderCallbacks<Cursor> {
   private SampleContactsAdapter adapter;
 
+  private static final String sReadContacts = Manifest.permission.READ_CONTACTS;
```
FabianTerhorst: sReadContacts -> READ_CONTACTS

picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
```diff
@@ -30,6 +34,9 @@
...
+  private static final int sReadContactsCode = 100;
```
FabianTerhorst: sReadContactsCode -> READ_CONTACTS_CODE

CoXier: Is it picasso's style ? It's different from [AOSP](https://source.android.com/source/code-style).

Ok I will change it.

FabianTerhorst: Constants are written in caps.

JakeWharton: AOSP's style is a historical accident and shouldn't be used anywhere except inside AOSP.

CoXier: Ok I get it.

CoXier: Is there anything for this pr to do?

CoXier:
Although this is a bug but it does not matter with this lib.

PR #1682, opened 2017-10-12
Migrate from Fest to Truth

### LINKED ISSUE #1592: Get rid of fest, fest-android. Replace with Truth.
PR welcome.

juandiana:
Fixes #1592.

They are very similar APIs. For the most part, only import changes and minor substitution of equivalent methods were needed (e.g. `containsOnly` for `containsExactly`, `contains(entry())` for `containsEntry()`).

Some notable bits:

- Assertions cannot be chained in Truth (design decision from the Truth team). Thus, chained assertions were split in multiple individual assertions.
- Fest's `Assertions.fail` was substituted with JUnit's `Assert.fail`.
- Truth’s `MapSubject` doesn’t have a `containsValue` method so I used `Map.containsValue` to assert its value.

**Note**: I had to upgrade the AGP from version 3.0-beta2 to 3.0-beta7 in order to work in AS 3.0 beta7 (current). Moreover, using this version of the AGP also required upgrading build tools from version 26.0.1 to 26.0.2.

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
+++ b/picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
@@ -389,8 +389,7 @@ public class BitmapHunterTest {
         dispatcher, cache, stats, action1);
     hunter.attach(action2);
     assertThat(hunter.getAction()).isEqualTo(action1);
-    assertThat(hunter.getActions()).hasSize(1);
-    assertThat(hunter.getActions()).contains(action2);
+    assertThat(hunter.getActions()).containsExactly(action2);
     assertThat(hunter.getPriority()).isEqualTo(HIGH);
   }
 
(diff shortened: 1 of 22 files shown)


picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -389,7 +389,8 @@
         dispatcher, cache, stats, action1);
     hunter.attach(action2);
     assertThat(hunter.getAction()).isEqualTo(action1);
-    assertThat(hunter.getActions()).hasSize(1).contains(action2);
+    assertThat(hunter.getActions()).hasSize(1);
+    assertThat(hunter.getActions()).contains(action2);
```
NightlyNexus: Could probably replace these hasSize+contains calls with containsExactly, but nbd, this is what the existing code had.

juandiana: Yeah, I agree it's cleaner to read this way. I actually tried to preserve everything as close as it was.
Made a new commit that replaces instances of this case.

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -1059,7 +1060,7 @@
       BitmapHunter.applyCustomTransformations(transformations, original);
       fail("Expected exception to be thrown.");
     } catch (RuntimeException e) {
-      assertThat(e).hasMessage("Transformation "
+      assertThat(e).hasMessageThat().isEqualTo("Transformation "
```
juandiana: The other case which would be suitable for changing is this one. Instead of using `isEqualTo`, we could use `contains`.

While I was going through I noticed that `hasMessageContaing` was being used in one case, but not on the rest (which use `hasMessage`). That's why I opted to use `isEqualTo` in order to preserve the original meaning.

JakeWharton:
This is awesome. Thank you!

PR #1690, opened 2017-10-23
Added support text as placeholder and error.

jordan1997:
I need the following function in my work-project when we show user avatar for example. May be other people have this problem too.


JakeWharton:
Supporting text is far too complicated. There's positioning, style, font, colors, etc. A Drawable gives you access to a Canvas where you can do whatever you want and that seems like the best way forward for text support.

jordan1997:
@JakeWharton will do

jordan1997:
Hi, @JakeWharton 
I do need suggestion from you, What better use standard Android gravity for positions text in the Drawable or create an enum for it?

PR #1698, opened 2017-11-01
Add TextDrawable

jordan1997:
The pull request implements use text as placeholder or error holder.
Implemented following functions:
1. Text color;
2. Text size;
3. Text font;
4. Background color;
5. Text gravity;
6. Text padding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ jniLibs
 build
 local.properties
 reports
+
+.idea/instapk.xml
+instapk.log*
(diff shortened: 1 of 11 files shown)


JakeWharton:
We do not want to be responsible for maintaining such complexity, and I don't see any reason to tie such functionality to this project since it can be used anywhere a Drawable can be used.

jordan1997:
@JakeWharton I will support with functionality in the future. 
About reason, developers will use this each time, when app shows provide with the avatar. (More apps show two letters in the image view when user doesn't have avatar).

jordan1997:
I fixed check style. @JakeWharton Can you reopen pull reqest? I want to update it.

JakeWharton:
We do not want to add this functionality to Picasso. If you'd like, you can
publish it as a standalone library for use with Picasso and anywhere else.

jordan1997:
Ok :(

PR #1726, opened 2018-01-10
Fix incorrect cache size reporting after Bitmap is recycled.

### LINKED ISSUE #1693: Fatal Exception: java.lang.IllegalStateException: com.squareup.picasso.LruCache.sizeOf() is reporting inconsistent results!
A lot of my users are getting the following error, and only on Android 8:

```
Fatal Exception: java.lang.IllegalStateException: com.squareup.picasso.LruCache.sizeOf() is reporting inconsistent results!
       at com.squareup.picasso.LruCache.trimToSize(LruCache.java:94)
       at com.squareup.picasso.LruCache.set(LruCache.java:84)
       at com.squareup.picasso.Dispatcher.performComplete(Dispatcher.java:370)
       at com.squareup.picasso.Dispatcher$DispatcherHandler.handleMessage(Dispatcher.java:502)
       at android.os.Handler.dispatchMessage(Handler.java:105)
       at android.os.Looper.loop(Looper.java:164)
       at android.os.HandlerThread.run(HandlerThread.java:65)
```

I have no idea how to reproduce it and don't know what's causing the error.

NightlyNexus:
Previously, if a user called recycle on a Bitmap in the cache, the allocation size would not be subtracted correctly (on SDK 25+) when the Bitmap was evicted from the cache.

I'm not sure that anyone should ever call recycle on a Bitmap in the cache, but it'd be difficult to communicate the error.

Closes #1693

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso/LruCacheTest.java
+++ b/picasso/src/test/java/com/squareup/picasso/LruCacheTest.java
@@ -218,9 +218,9 @@ public class LruCacheTest {
 
   private void assertSnapshot(LruCache cache, Object... keysAndValues) {
     List<Object> actualKeysAndValues = new ArrayList<>();
-    for (Map.Entry<String, Bitmap> entry : cache.map.entrySet()) {
+    for (Map.Entry<String, LruCache.BitmapAndSize> entry : cache.map.entrySet()) {
       actualKeysAndValues.add(entry.getKey());
-      actualKeysAndValues.add(entry.getValue());
+      actualKeysAndValues.add(entry.getValue().bitmap);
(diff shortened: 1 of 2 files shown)


JakeWharton:
Oof

picasso/src/test/java/com/squareup/picasso/LruCacheTest.java
```diff
@@ -17,6 +17,7 @@
 package com.squareup.picasso;
 
 import android.graphics.Bitmap;
+import android.util.Pair;
```
JakeWharton: Unused?

PR #1743, opened 2018-02-28
Add Robolectric configs.

NightlyNexus:
I've always had problems getting these Robolectric tests working locally.
Today, the error was `No 'constants' field in @Config annotation!` and it went away after adding these configs. (Config could be added in a subclass of RobolectricGradleTestRunner, but who wants more Robolectric code?) 

Close if it's not a problem for anybody else, and I can patch locally.

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso/AssetRequestHandlerTest.java
+++ b/picasso/src/test/java/com/squareup/picasso/AssetRequestHandlerTest.java
@@ -8,11 +8,14 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.robolectric.RobolectricGradleTestRunner;
+import org.robolectric.annotation.Config;
 
+import static android.os.Build.VERSION_CODES.M;
 import static com.google.common.truth.Truth.assertThat;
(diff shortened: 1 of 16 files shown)


picasso/src/test/java/com/squareup/picasso/AssetRequestHandlerTest.java
```diff
@@ -8,11 +8,14 @@
...
 import org.mockito.Mock;
 import org.robolectric.RobolectricGradleTestRunner;
+import org.robolectric.annotation.Config;
 
+import static android.os.Build.VERSION_CODES.M;
 import static com.google.common.truth.Truth.assertThat;
 import static org.mockito.MockitoAnnotations.initMocks;
 
 @RunWith(RobolectricGradleTestRunner.class)
+@Config(constants = BuildConfig.class, sdk = M)
```
vanniktech: You don't need the BuildConfig class when using AGP 3.0.0

JakeWharton: Plus we're about to stop generating it...

jrodbx:
Closing.  This works fine locally for me and on CI.

NightlyNexus:
👍

PR #1744, opened 2018-02-28
Disable BuildConfig generation.

NightlyNexus:


NightlyNexus:
failed, hm? works locally for me at least after also applying #1743. have no idea what's wrong in Robolectric...

NightlyNexus:
okay, errors make sense now. just missing the BuildConfig that Robolectric wants.

JakeWharton:
Rebase? Does it still fail?

NightlyNexus:
tests still fail for me locally after rebasing.
`java.lang.RuntimeException: Could not load class: com.squareup.picasso.BuildConfig`

vanniktech:
Afaik, robolectric will try load it.

JakeWharton:
There's a lever we can pull somewhere to get it to stop.

jrodbx:
IIRC, this requires bumping to a later version of Robolectric which will introduce other issues (see https://github.com/robolectric/robolectric/issues/3242 for an example).  IMO, it's not worth the effort at the moment.

PR #1754, opened 2018-03-03
Gradle cleanup

jrodbx:


picasso-pollexor/build.gradle
```diff
@@ -21,22 +19,18 @@ android {
   }
 
   dexOptions {
-    preDexLibraries = !rootProject.ext.isCi
-  }
-
-  defaultConfig {
-    minSdkVersion rootProject.ext.minSdkVersion
+    preDexLibraries = !isCi
```
JakeWharton: This isn't needed anymore

build.gradle
```diff
@@ -1,10 +1,41 @@
...
+        okhttp             : "com.squareup.okhttp3:okhttp:${versions.okhttp}",
+        mockWebServer      : "com.squareup.okhttp3:mockwebserver:${versions.okhttp}",
+        pollexor           : 'com.squareup:pollexor:2.0.4',
+        supportV4          : "com.android.support:support-v4:${versions.supportLibrary}",
+        supportAnnotations : "com.android.support:support-annotations:${versions.supportLibrary}",
+        supportExifInterface : "com.android.support:exifinterface:${versions.supportLibrary}",
+        junit              : 'junit:junit:4.12',
+        truth              : 'com.google.truth:truth:0.36',
+        robolectric        : 'org.robolectric:robolectric:3.1',
+        mockito            : 'org.mockito:mockito-core:1.9.5'
```
JakeWharton: Get rid of alignment? It's screwed up anyway

picasso-sample/build.gradle
```diff
@@ -33,18 +30,15 @@ android {
     preDexLibraries = !rootProject.ext.isCi
```
JakeWharton: Delete

picasso-sample/build.gradle
```diff
@@ -33,18 +30,15 @@ android {
...
-  defaultConfig {
-    minSdkVersion rootProject.ext.minSdkVersion
-    targetSdkVersion rootProject.ext.targetSdkVersion
-    applicationId 'com.example.picasso'
-
-    versionCode 1
-    versionName '1.0'
-  }
-
   buildTypes {
```
JakeWharton: Whaaaaaa?

Delete block.

build.gradle
```diff
@@ -1,10 +1,41 @@
...
+        'okhttp': '3.10.0',
+        'supportLibrary': '27.1.0'
+    ]
+
+  ext.deps = [
+        androidPlugin      : 'com.android.tools.build:gradle:3.0.1',
+        okhttp             : "com.squareup.okhttp3:okhttp:${versions.okhttp}",
+        mockWebServer      : "com.squareup.okhttp3:mockwebserver:${versions.okhttp}",
+        pollexor           : 'com.squareup:pollexor:2.0.4',
+        supportV4          : "com.android.support:support-v4:${versions.supportLibrary}",
```
JakeWharton: Also support-v4 isn't a thing anymore. We should depend on a granular library. What's this used for?

picasso/build.gradle
```diff
@@ -21,26 +20,20 @@ android {
...
 
   dexOptions {
-    preDexLibraries = !rootProject.ext.isCi
-  }
-
-  defaultConfig {
-    minSdkVersion rootProject.ext.minSdkVersion
-
-    consumerProguardFiles 'consumer-proguard-rules.txt'
+    preDexLibraries = !isCi
```
JakeWharton: Delete

picasso/build.gradle
```diff
@@ -21,26 +20,20 @@ android {
...
-  api dep.supportAnnotations
-  implementation dep.supportExifInterface
-  testImplementation dep.junit
-  testImplementation dep.truth
-  testImplementation dep.supportV4
-  testImplementation dep.robolectric
-  testImplementation dep.mockito
-  testImplementation dep.mockWebServer
+  api deps.okhttp
+  compileOnly deps.supportAnnotations
```
JakeWharton: Why this and not `implementation`?

build.gradle
jrodbx: android.support.v4.app.NotificationCompat
android.support.v4.app.FragmentActivity
android.support.v4.widget.CursorAdapter
android.support.v4.app.LoaderManager
android.support.v4.content.CursorLoader
android.support.v4.content.Loader
android.support.v4.app.Fragment

JakeWharton: so compat and fragment? that's not too bad.

jrodbx: I removed the dependency from the `picasso` module.

picasso/build.gradle
jrodbx: the annotations aren't needed for runtime, so figured why not remove it from the runtime classpath?

JakeWharton: They're transitive of the exif library already. Without that, in the insane case a user doesn't have any other dependencies pulling annotations (basically impossible) it would cause ProGuard to fail. I would just leave it as implementation.

PR #1755, opened 2018-03-03
Gradle cleanup

jrodbx:
Supercedes https://github.com/square/picasso/pull/1754


.travis.yml
```diff
@@ -1,10 +1,5 @@
 language: android
 
-android:
```
NightlyNexus: ah, i misunderstood. thanks for cleaning this up.

PR #1758, opened 2018-03-03
Use Android's LruCache.

swankjesse:


picasso/src/main/java/com/squareup/picasso/LruCache.java
```diff
@@ -19,159 +19,99 @@
...
   }
 
   @Override public void set(@NonNull String key, @NonNull Bitmap bitmap) {
     if (key == null || bitmap == null) {
       throw new NullPointerException("key == null || bitmap == null");
     }
 
-    int addedSize = Utils.getBitmapBytes(bitmap);
-    if (addedSize > maxSize) {
-      return;
```
swankjesse: this was a theoretical bug . . . replacing an existing key didn’t evict if the size was too big

JakeWharton: Are you sure? It was contributed because it was observed behavior.

JakeWharton: Oh sorry, I'm looking at the wrong method.

JakeWharton: Wait no I'm not... and you retained the behavior but added a comment?

swankjesse: No, the “putting a giant image clears cache” is totally a real problem.

But there’s a theoretical problem with the fix. You have an image that’s beneath the limit in the cache. You try to update that with a higher-bytes copy of the same image. That doesn’t evict the lower-bytes predecessor, though it should.

JakeWharton: Ah, cool. 👍

picasso/src/main/java/com/squareup/picasso/LruCache.java
```diff
@@ -19,159 +19,99 @@
...
-        value = toEvict.getValue();
-        map.remove(key);
-        size -= value.allocationSize;
-        evictionCount++;
-      }
+    // If the bitmap is too big for the cache, don't even attempt to store it. Doing so will cause
+    // the cache to be cleared. Instead just evict an existing element with the same key if it
+    // exists.
+    if (byteCount > maxSize()) {
+      cache.remove(key);
```
NightlyNexus: nice, thank you. was my bug from 2 years ago. i'll update with a test.

NightlyNexus:
I think it does want to be a public type?
People might want to configure a different size for their memory caches.
The constructor that accepts a Context seems like it shouldn't be public, though.

swankjesse:
Probably shouldn't be a public type. I think users should configure the cache size in bytes and we should do the eviction policy.

NightlyNexus:
Oh, I gotcha.
Don't expose Cache at all? Let users set the cache size, use the default, or set no cache when building the Picasso.

PR #1760, opened 2018-03-03
Add test for evicting old image when new large one can't fit in the c…

NightlyNexus:
…ache.

https://github.com/square/picasso/pull/1758#discussion_r172005297


JakeWharton:
Nice test

PR #1762, opened 2018-03-03
Drop JUnit assertEquals calls in favor of Truth assertThat.

vanniktech:


picasso/src/test/java/com/squareup/picasso/LruCacheTest.java
```diff
@@ -194,7 +193,7 @@
     cache.set("25", size25);
     assertHit(cache, "16", size16);
     assertMiss(cache, "25");
-    assertEquals(cache.size(), 16);
+    assertThat(cache.size()).isEqualTo(16);
```
vanniktech: This and the one below also had the parameters in the wrong order.

NightlyNexus: Good catch, thx.

PR #1764, opened 2018-03-05
Add an Okio replacement for MarkableInputStream.

swankjesse:
I'll wire it up in a follow-up PR.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/SourceMarker.java
+++ b/picasso/src/main/java/com/squareup/picasso/SourceMarker.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.picasso;
+
+import java.io.IOException;
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.ForwardingSource;
+import okio.Okio;
+import okio.Source;
+
+/**
+ * Builds a buffered source that can rewind to a marked position earlier in the stream.
+ *
+ * <p>Mark potential positions to rewind back to with {@link #mark}; rewind back to these positions
+ * with {@link #reset}. Both operations apply to the position in the {@linkplain #source() buffered
+ * source}; resetting will impact the buffer.
+ *
+ * <p>When marking it is necessary to specify how much data to retain. Once you advance above this
+ * limit, the mark is discarded and resetting is not permitted. This may be used to lookahead a
+ * fixed number of bytes without loading an entire stream into memory. To reset an arbitrary
+ * number of bytes use {@code mark(Long#MAX_VALUE)}.
+ */
+public final class SourceMarker {
+
+  /*
+   * This class wraps the underlying source in a MarkSource to support mark and reset. It creates a
+   * BufferedSource for the caller so that it can track its offsets and manipulate its buffer.
+   */
+
+  /**
+   * The offset into the underlying source. To compute the user's offset start with this and
+   * subtract userBuffer.size().
+   */
+  long offset;
+
+  /** The offset of the earliest mark, or -1 for no mark. */
+  long mark = -1L;
+
+  /** The offset of the latest readLimit, or -1 for no mark. */
+  long limit = -1L;
+
+  boolean closed;
+
+  final MarkSource markSource;
+  final BufferedSource userSource;
+
+  /** A copy of the underlying source's data beginning at {@code mark}. */
+  final Buffer markBuffer;
+
+  /** Just the userSource's buffer. */
+  final Buffer userBuffer;
+
+  public SourceMarker(Source source) {
+    this.markSource = new MarkSource(source);
+    this.markBuffer = new Buffer();
+    this.userSource = Okio.buffer(markSource);
+    this.userBuffer = userSource.buffer();
+  }
+
+  public BufferedSource source() {
+    return userSource;
+  }
+
+  /**
+   * Marks the current position in the stream as one to potentially return back to. Returns the
+   * offset of this position. Call {@link #reset(long)} with this position to return to it later. It
+   * is an error to call {@link #reset(long)} after consuming more than {@code readLimit} bytes from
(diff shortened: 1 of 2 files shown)


picasso/src/test/java/com/squareup/picasso/SourceMarkerTest.java
```diff
@@ -0,0 +1,251 @@
...
+          .isEqualTo("cannot reset to 3: out of range");
+    }
+  }
+
+  @Test public void exceedLimitTest() throws Exception {
+    SourceMarker marker = new SourceMarker(new Buffer().writeUtf8("ABCDEFGHIJKLMNOPQRSTUVWXYZ"));
+    BufferedSource source = marker.source();
+
+    assertThat(source.readUtf8(3)).isEqualTo("ABC");
+    long pos3 = marker.mark(Long.MAX_VALUE); // DEFGHIJ
```
NightlyNexus: is this comment right? copy-paste accident?

picasso/src/test/java/com/squareup/picasso/SourceMarkerTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    try {
+      marker.mark(-1L);
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessageThat().isEqualTo("readLimit < 0: -1");
+    }
+  }
+
+  @Test public void resetAfterEof() throws Exception {
```
NightlyNexus: definitely cool.

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
```diff
@@ -0,0 +1,185 @@
...
+ * <p>Mark potential positions to rewind back to with {@link #mark}; rewind back to these positions
+ * with {@link #reset}. Both operations apply to the position in the {@linkplain #source() buffered
+ * source}; resetting will impact the buffer.
+ *
+ * <p>When marking it is necessary to specify how much data to retain. Once you advance above this
+ * limit, the mark is discarded and resetting is not permitted. This may be used to lookahead a
+ * fixed number of bytes without loading an entire stream into memory. To reset an arbitrary
+ * number of bytes use {@code mark(Long#MAX_VALUE)}.
+ */
+public final class SourceMarker {
```
NightlyNexus: not public for Picasso, yeah?

jrodbx:
I'll take a look in the morning!

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
```diff
@@ -0,0 +1,185 @@
...
+    return userSource;
+  }
+
+  /**
+   * Marks the current position in the stream as one to potentially return back to. Returns the
+   * offset of this position. Call {@link #reset(long)} with this position to return to it later. It
+   * is an error to call {@link #reset(long)} after consuming more than {@code readLimit} bytes from
+   * {@linkplain #source() the source}.
+   */
+  public long mark(long readLimit) throws IOException {
```
JakeWharton: I don't think `readLimit` is useful here. We don't know how much the image decoder will consume in order to discern the size of the image. Picasso will always pass `Long.MAX_VALUE` to this to allow reading unlimited bytes up to EOF.

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
```diff
@@ -0,0 +1,185 @@
...
+import okio.ForwardingSource;
+import okio.Okio;
+import okio.Source;
+
+/**
+ * Builds a buffered source that can rewind to a marked position earlier in the stream.
+ *
+ * <p>Mark potential positions to rewind back to with {@link #mark}; rewind back to these positions
+ * with {@link #reset}. Both operations apply to the position in the {@linkplain #source() buffered
+ * source}; resetting will impact the buffer.
```
JakeWharton: The implementation can be simplified to remove this functionality and have it always reset back to 0 if you'd like. We'll never pass anything but 0

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
```diff
@@ -0,0 +1,185 @@
...
+        if (result == -1L) return -1L;
+        offset += result;
+        return result;
+      }
+
+      // If we can read from markBuffer, do that.
+      if (offset < mark + markBuffer.size()) {
+        long posInBuffer = offset - mark;
+        long result = Math.min(byteCount, markBuffer.size() - posInBuffer);
+        markBuffer.copyTo(sink, posInBuffer, result);
```
JakeWharton: We'll only be consuming this as an `InputStream` so it makes me sad that this will be doing double copies for everything.

picasso/src/test/java/com/squareup/picasso/SourceMarkerTest.java
```diff
@@ -0,0 +1,251 @@
...
+    long pos10k = marker.mark(15_000);
+    assertThat(source.readUtf8(10_000)).isEqualTo(bs);
+    long pos20k = marker.mark(15_000);
+    assertThat(source.readUtf8(10_000)).isEqualTo(cs);
+    marker.reset(pos20k);
+    marker.reset(pos10k);
+    assertThat(source.readUtf8(30_000)).isEqualTo(bs + cs + ds);
+  }
+
+  private String repeat(char c, int count) {
```
JakeWharton: Java 11 has a `String.repeat(int count)` method so come september we can kill all these.

picasso/src/test/java/com/squareup/picasso/SourceMarkerTest.java
swankjesse: Yup. Fixed!

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
swankjesse: Yep, that’s a mistake. Fixed.

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
swankjesse: Agreed for Picasso. I think I wanna promote this into Okio samples, or perhaps elsewhere.

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
swankjesse: I think SourceMarker might be the wrong abstraction for Picasso. Instead it just wants something that buffers the stream, kind of like OkHttp’s peekBody(). Lemme move this to Okio and get something simpler for Picasso.

picasso/src/main/java/com/squareup/picasso/SourceMarker.java
swankjesse: Will propose something better-suited to Picasso’s use.

picasso/src/test/java/com/squareup/picasso/SourceMarkerTest.java
swankjesse: Yay.

PR #1767, opened 2018-03-05
Cleanup samples

jrodbx:
* Remove obsolete manifest entries
* Fix Contacts activity to request permissions on M+
* Fix Notifications to use channels on O+
* Fix notification content text on API 22

### DIFF
--- a/picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
+++ b/picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
@@ -16,18 +16,26 @@
  */
 package com.example.picasso;
 
+import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.net.Uri;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.v4.app.ActivityCompat;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
 import android.support.v4.content.Loader;
 import android.widget.ListView;
+import android.widget.Toast;
 
+import static android.Manifest.permission.READ_CONTACTS;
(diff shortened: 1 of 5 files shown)


picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
```diff
@@ -40,9 +48,34 @@
...
   }
 
+  @Override public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions,
+      @NonNull int[] grantResults) {
+    if (requestCode == REQUEST_READ_CONTACTS) {
+      if (grantResults.length > 0
+          && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+        loadContacts();
+      } else {
+        Toast.makeText(getApplicationContext(), "Read contacts permission denied",
```
NightlyNexus: to nitpick: just use `this` instead of the application.

jrodbx: wouldn't that cause a leak, since the toast could outlive the activity, say, by hitting the back button?

NightlyNexus: ah, i guess so. thought Toast already called getApplicationContext, but I'm wrong!

JakeWharton: Toasts immediately leave the process. Using `this` is fine.

JakeWharton:
Close the other PRs that do this when you merge.

picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
jrodbx: Updated

PR #1769, opened 2018-03-05
Remove Cache from public API.

NightlyNexus:
Users can set the max byte count, use the default, or specify no memory caching.

Here's a proposal for an API change. If this sounds good, an implementation follow-up might be removing `Cache` in favor of passing around a nullable Cache/LruCache.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Cache.java
+++ b/picasso/src/main/java/com/squareup/picasso/Cache.java
@@ -24,7 +24,7 @@ import android.graphics.Bitmap;
  * your {@link Cache} implementation is thread safe when {@link Cache#get(String)} or {@link
  * Cache#set(String, android.graphics.Bitmap)} is called.
  */
-public interface Cache {
+interface Cache {
   /** Retrieve an image for the specified {@code key} or {@code null}. */
   Bitmap get(String key);
 
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -768,14 +768,18 @@ public class Picasso {
     }
 
     /** Specify the memory cache used for the most recent images. */
-    public Builder memoryCache(@NonNull Cache memoryCache) {
-      if (memoryCache == null) {
-        throw new IllegalArgumentException("Memory cache must not be null.");
+    public Builder withCacheSize(int maxByteCount) {
+      if (maxByteCount < 0) {
+        throw new IllegalArgumentException("maxByteCount < 0: " + maxByteCount);
       }
-      if (this.cache != null) {
+      if (cache != null) {
         throw new IllegalStateException("Memory cache already set.");
       }
-      this.cache = memoryCache;
+      if (maxByteCount == 0) {
+        cache = Cache.NONE;
+      } else {
+        cache = new LruCache(maxByteCount);
+      }
       return this;
     }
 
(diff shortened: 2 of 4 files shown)


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -768,14 +768,22 @@ public Builder executor(@NonNull ExecutorService executorService) {
...
         throw new IllegalStateException("Memory cache already set.");
       }
-      this.cache = memoryCache;
+      this.cache = new LruCache(maxByteCount);
+      return this;
+    }
+
+    public Builder noMemoryCache() {
+      if (this.cache != null) {
+        throw new IllegalStateException("Memory cache already set.");
```
NightlyNexus: left this around for now, but I'd like to remove these checks for a newBuilder nicety.

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -768,14 +768,22 @@ public Builder executor(@NonNull ExecutorService executorService) {
...
       }
       if (this.cache != null) {
         throw new IllegalStateException("Memory cache already set.");
       }
-      this.cache = memoryCache;
+      this.cache = new LruCache(maxByteCount);
+      return this;
+    }
+
+    public Builder noMemoryCache() {
```
NightlyNexus: don't like the names: noMemoryCache vs memoryCacheBytes (prefix vs. suffix)

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -768,14 +768,22 @@ public Builder executor(@NonNull ExecutorService executorService) {
     }
 
     /** Specify the memory cache used for the most recent images. */
-    public Builder memoryCache(@NonNull Cache memoryCache) {
-      if (memoryCache == null) {
-        throw new IllegalArgumentException("Memory cache must not be null.");
+    public Builder memoryCacheBytes(int maxByteCount) {
```
jrodbx: `withCacheSize(int)`?

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -768,14 +768,22 @@ public Builder executor(@NonNull ExecutorService executorService) {
...
+      if (maxByteCount <= 0) {
+        throw new IllegalArgumentException("maxByteCount <= 0: " + maxByteCount);
```
jrodbx: < 0, illegal, but == 0, Cache.NONE?

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: depending on thoughts above (re: withCacheSize(0)), this could go away.

picasso/src/main/java/com/squareup/picasso/Picasso.java
NightlyNexus: yes. dig it. and then 0 is no cache. thanks.

picasso/src/main/java/com/squareup/picasso/Cache.java
```diff
@@ -24,7 +24,7 @@
  * your {@link Cache} implementation is thread safe when {@link Cache#get(String)} or {@link
  * Cache#set(String, android.graphics.Bitmap)} is called.
  */
-public interface Cache {
+interface Cache {
```
JakeWharton: Can we lose this in a follow-up and rename `LruCache` to `Cache`?

gubatron:
found it :)

picasso/src/main/java/com/squareup/picasso/Cache.java
NightlyNexus: Yes, that's exactly what I was thinking.

NightlyNexus:
Will need a Picasso.clearMemoryCache(), I think.

PR #1770, opened 2018-03-05
Extract global Picasso singleton into separate optional artifact

jrodbx:
To manage global singletons, consumers can either use dependency injection or include an optional `picasso-provider` dependency.


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -166,16 +162,66 @@
...
+   * available on API 14+ <em>or</em> if you are using a standalone library that provides a disk
+   * cache on all API levels like OkHttp)</li>
+   * <li>Three download threads for disk and network access.</li>
+   * </ul>
+   * <p>
+   * If these settings do not meet the requirements of your application, you can construct your own
+   * with full control over the configuration by using {@link Picasso.Builder} to create a
+   * customized {@link Picasso} instance.
+   */
+  public Picasso(Context context) {
```
NightlyNexus: Why the extra constructor? You can always do `new Picasso.Builder(context).build()`.
I worry about people who will fire and forget. `new Picasso(context).load().into()`. I'd hope the awkwardness of the builder causes pause; you only have to do it once in the common case!

jrodbx: Took a page out of OkHttp and how OkHttpClient is modeled.  Fair point though; let's see what the others think.

picasso-sample/src/main/java/com/example/picasso/SampleApplication.java
```diff
@@ -0,0 +1,19 @@
+package com.example.picasso;
+
+import android.app.Application;
+import com.squareup.picasso.Picasso;
+
+// Note: Production apps should instead use the dependency injection design pattern to provide
```
JakeWharton: You're already using the DI pattern everywhere below this. I think you should say DI _library_ or traditional singleton pattern here.

picasso/src/main/java/com/squareup/picasso/Picasso.java
JakeWharton: I would prefer having just the builder because it keeps defaults in one location.

picasso-sample/src/main/java/com/example/picasso/SampleApplication.java
jrodbx: updated.  see `PicassoProvider.get(Context)`.

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: removed the proposed added constructor.

picasso-provider/src/main/java/com/squareup/picasso/provider/PicassoProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * available on API 14+ <em>or</em> if you are using a standalone library that provides a disk
+ * cache on all API levels like OkHttp)</li>
+ * <li>Three download threads for disk and network access.</li>
+ * </ul>
+ * <p>
+ * If these settings do not meet the requirements of your application, you can construct your own
+ * with full control over the configuration by using {@link Picasso.Builder} to create a
+ * customized {@link Picasso} instance.
+ *
+ * Note: Production apps could instead use dependency injection to provide their Picasso instances.
```
JakeWharton: Need `<p>` or this will be combined into previous paragraph.

picasso-provider/src/main/java/com/squareup/picasso/provider/PicassoProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public final class PicassoProvider {
+  @SuppressLint("StaticFieldLeak")
+  private static Picasso instance;
+
+  public static Picasso get(Context context) {
+    if (instance == null) {
+      synchronized (Picasso.class) {
```
JakeWharton: `PicassoProvider.class`

picasso-provider/src/main/java/com/squareup/picasso/provider/PicassoProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * with full control over the configuration by using {@link Picasso.Builder} to create a
+ * customized {@link Picasso} instance.
+ *
+ * Note: Production apps could instead use dependency injection to provide their Picasso instances.
+ */
+public final class PicassoProvider {
+  @SuppressLint("StaticFieldLeak")
+  private static Picasso instance;
+
+  public static Picasso get(Context context) {
```
JakeWharton: Should we use the content provider trick here since we don't let you customize the instance?

settings.gradle
```diff
@@ -1,3 +1,4 @@
+include ':picasso-provider'
```
JakeWharton: Should be between `pollexor` and `sample`

picasso-provider/src/main/java/com/squareup/picasso/provider/PicassoProvider.java
jrodbx: Good catch.  Updated.

picasso-provider/src/main/java/com/squareup/picasso/provider/PicassoProvider.java
jrodbx: Sure.  Reintroduced.

settings.gradle
jrodbx: Updated.

picasso-provider/src/main/java/com/squareup/picasso/provider/PicassoProvider.java
jrodbx: Added.

NightlyNexus:
nice.

PR #1773, opened 2018-03-05
Add required internet permission.

NightlyNexus:


JakeWharton:
We never did this because nothing prevents Picasso's use without internet.

NightlyNexus:
ooooh, good point!

PR #1775, opened 2018-03-06
Fix test Bitmap.Config.

NightlyNexus:
This parameter is not nullable.

It's for a change to yank out the Cache abstraction.
Surprisingly, it doesn't look the Android SDK checks for a null input and would only fail later. I'll file a bug when I get a minute.


NightlyNexus:
and I'm wrong. a null config will throw eventually in the createBitmap call. just the Robolectric code doesn't check.

PR #1778, opened 2018-03-06
Remove internal Cache abstraction.

NightlyNexus:
Make a single nullable cache type, with null signifying no cache.

depends on the evictAll PR.

I was hoping to make only an implementation change, but it broke all the tests. Many of the tests are only broken because I made `Cache` final, so I could break it into another PR without final if that's easier to review.


picasso/src/main/java/com/squareup/picasso/Cache.java
```diff
@@ -15,56 +15,105 @@
...
-  Bitmap get(String key);
-
-  /** Store an image in the cache for the specified {@code key}. */
-  void set(String key, Bitmap bitmap);
+import static com.squareup.picasso.Utils.KEY_SEPARATOR;
 
-  /** Returns the current size of the cache in bytes. */
-  int size();
+/** A memory cache which uses a least-recently used eviction policy. */
+final class Cache {
```
NightlyNexus: this diff is a mess, but this is a pure copy-paste from LruCache.

picasso/src/main/java/com/squareup/picasso/Cache.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013 Square, Inc.
+ * Copyright (C) 2011 The Android Open Source Project
```
NightlyNexus: this is out of date now, anyway, after Jesse's change to use Android's LruCache.
Should we do `Copyright (C) 2018 Square, Inc.`?

picasso/src/main/java/com/squareup/picasso/Cache.java
JakeWharton: I think we should call this `PlatformLruCache` in case we decide to re-add the `Cache` interface in the future it will prevent another rename. I'm not 100% convinced that pluggable caches are a bad thing, I just think we might as well hide it for now until a clear use case arises.

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -717,13 +732,14 @@ public Builder withCacheSize(int maxByteCount) {
       if (maxByteCount < 0) {
         throw new IllegalArgumentException("maxByteCount < 0: " + maxByteCount);
       }
-      if (cache != null) {
+      if (cacheSet) {
         throw new IllegalStateException("Memory cache already set.");
       }
+      cacheSet = true;
```
JakeWharton: This seems to match nullability of the field. Can we just use that instead like before?

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -86,7 +87,7 @@
 
   @Mock Context context;
   @Mock Picasso picasso;
-  @Mock Cache cache;
```
JakeWharton: One down, hundreds to go...

picasso/src/test/java/com/squareup/picasso/NetworkRequestHandlerTest.java
```diff
@@ -67,7 +68,7 @@
   };
 
   @Mock Picasso picasso;
-  @Mock Cache cache;
+  final @Nullable Cache cache = null;
```
JakeWharton: Hmm just inline?

picasso/src/main/java/com/squareup/picasso/Picasso.java
NightlyNexus: i can get rid of the legacy "already set" message.
but, unfortunately not this flag because i'm overloading the meaning of null.

in `build`, we have to know whether a user explicitly passed in `0` for a null/no cache or whether we have the unset value and should create the default cache for the user.

picasso/src/test/java/com/squareup/picasso/NetworkRequestHandlerTest.java
NightlyNexus: yes.

picasso/src/main/java/com/squareup/picasso/Cache.java
NightlyNexus: sounds good.

picasso/src/main/java/com/squareup/picasso/RequestCreator.java
```diff
@@ -409,7 +409,7 @@ public RequestCreator noFade() {
    * Synchronously fulfill this request. Must not be called from the main thread.
    * <p>
    * <em>Note</em>: The result of this operation is not cached in memory because the underlying
-   * {@link Cache} implementation is not guaranteed to be thread-safe.
+   * {@link PlatformLruCache} implementation is not guaranteed to be thread-safe.
```
NightlyNexus: untrue.

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -745,13 +754,11 @@ public Builder withCacheSize(int maxByteCount) {
...
-      if (cache != null) {
-        throw new IllegalStateException("Memory cache already set.");
-      }
+      cacheSet = true;
       if (maxByteCount == 0) {
-        cache = Cache.NONE;
+        cache = null;
       } else {
-        cache = new LruCache(maxByteCount);
+        cache = new PlatformLruCache(maxByteCount);
```
jrodbx: How about always creating a cache (even of size 0), instead of making the cache nullable and requiring null checks everywhere?  Adds less noise to this change.

picasso/src/main/java/com/squareup/picasso/RequestCreator.java
jrodbx: Wanna update/remove this comment?

picasso/src/main/java/com/squareup/picasso/Picasso.java
NightlyNexus: I considered it but was slightly concerned about overhead in ever insertion call for no cache.
That said, it's rare somebody doesn't want a cache.

Actually, could we cheat by checking maxSize and no-oping the cache `set` call?

JakeWharton: I don't think it's worth it unless we measure it to be a problem.

NightlyNexus: yep, sounds good.
i was hoping you could do skip the cache with a RequestTransformer that changes the memory policy on all requests, but you can't.

anyway, will update tonight.

picasso/src/main/java/com/squareup/picasso/RequestCreator.java
NightlyNexus: yep, can do.

picasso/src/main/java/com/squareup/picasso/Picasso.java
NightlyNexus: OH SHOOT. android.util.LruCache doesn't let us have 0 for the size.

NightlyNexus: went with making a 1-byte cache. we can always wait for somebody to complain if it breaks a use case.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -751,14 +751,7 @@ public Builder withCacheSize(int maxByteCount) {
...
       }
-      if (cache != null) {
-        throw new IllegalStateException("Memory cache already set.");
-      }
-      if (maxByteCount == 0) {
-        cache = Cache.NONE;
-      } else {
-        cache = new LruCache(maxByteCount);
-      }
+      cache = new PlatformLruCache(maxByteCount != 0 ? maxByteCount : 1);
```
JakeWharton: Let's move this ternary inside `PlatformLruCache` as it's an implementation detail of that type.

NightlyNexus: Will do. Will update the test with 0, then, too.

NightlyNexus: updated and updated tests. I didn't add a test for this in PlatformLruCacheTest, since it's not desired behavior, exactly. Let me know if I should!

picasso/src/test/java/com/squareup/picasso3/PlatformLruCacheTest.java
```diff
@@ -93,14 +93,14 @@
 
   @Test public void constructorDoesNotAllowZeroCacheSize() {
```
JakeWharton: Delete

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: just curious: what's not desired behavior?  either way, sgtm.

NightlyNexus: I mean that I could have had a test showing that a 0-size cache can really hold a 1sqr alpha_8. But, that's not desired behavior. It'd be a test of a workaround.

jrodbx: ah got it.

PR #1784, opened 2018-03-07
Require all into calls to be on the main thread.

NightlyNexus:


picasso/src/main/java/com/squareup/picasso/RequestCreator.java
```diff
@@ -588,6 +588,7 @@ public void into(@NonNull RemoteViews remoteViews, @IdRes int viewId, int notifi
    */
   public void into(@NonNull RemoteViews remoteViews, @IdRes int viewId, int notificationId,
       @NonNull Notification notification, @Nullable String notificationTag, Callback callback) {
+    checkMain();
```
JakeWharton: Aren't these calls typically made from a background thread?

NightlyNexus: updating notifications? i'd think they're normally from BroadcastReceiver.onReceive or a foreground Service.

that said, there's no reason for this change. i think the `Target` overload could remove its check, too, if anybody ever asks for it.

PR #1788, opened 2018-03-07
Update builder doc for cache size.

NightlyNexus:
i'm thinking it should maybe be `cacheSize` cuz we don't prefix the other setters like `client` with `with`.
nitpicks for a later time.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso/Picasso.java
@@ -719,7 +719,10 @@ public class Picasso {
       return this;
     }
 
-    /** Specify the memory cache used for the most recent images. */
+    /**
+     * Specify the memory cache size in bytes to use for the most recent images,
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -719,7 +719,10 @@ public Builder executor(@NonNull ExecutorService executorService) {
       return this;
     }
 
-    /** Specify the memory cache used for the most recent images. */
+    /**
+     * Specify the memory cache size in bytes to use for the most recent images,
+     * or 0 for no memory caching.
+     */
```
jrodbx:
```
Specify the memory cache size in bytes to use for the most recent images.
A size of 0 disables in-memory caching.
```

NightlyNexus: yeah.

Issue #1800, opened 2018-03-09
Let RequestTransformers change network and memory policy.


NightlyNexus:
Disallow changing to no memory cache lookup? Make all cache lookups before creating Requests?

I have something like this working and also fixed #1797, but I have to clean it up when I have time tonight.
Feel free to tell me if it's a bad idea haha.

JakeWharton:
With #739 we should just make them like interceptors and provide the ability to call `.newBuilder()` on a `Request`.

jrodbx:
Seems like this should remain a global setting in the Picasso object itself, but I might be missing something, but what's the vision here?  In other words, as a request travels through the transformer pipeline, what conditions would lend one to change a network or memory policy of NO_CACHE to NO_STORE or OFFLINE?

NightlyNexus:
These are properties of a request, so it seems like the should be transferable with newBuilder().

but maybe we leave them as properties for internal use and don't expose them for the RequestTransformers? unsure.

PR #1801, opened 2018-03-09
Introduce RequestTransformer chaining

jrodbx:


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -471,14 +472,23 @@ public void shutdown() {
...
   Request transformRequest(Request request) {
-    Request transformed = requestTransformer.transformRequest(request);
-    if (transformed == null) {
-      throw new IllegalStateException("Request transformer "
-          + requestTransformer.getClass().getCanonicalName()
-          + " returned null for "
-          + request);
+    if (requestTransformers == null) {
+      return request;
     }
```
jrodbx: I wrestled between this and an empty ArrayList().  Effective Java says the latter, but Picasso has a history of conservative memory usage.  Side effect: test setup is cleaner.  Thoughts?

NightlyNexus: Collections.emptyList() has no memory effect.

NightlyNexus: oh, i see. in the builder. yeah, null seems fine.

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -768,20 +778,18 @@ public Builder listener(@NonNull Listener listener) {
...
+    /** Add a transformer that observes and potentially modify all incoming requests. */
+    public Builder addRequestTransformer(RequestTransformer transformer) {
       if (transformer == null) {
         throw new IllegalArgumentException("Transformer must not be null.");
       }
-      if (this.transformer != null) {
+      if (requestTransformers == null) {
+        requestTransformers = new ArrayList<>();
+      }
+      if (requestTransformers.contains(transformer)) {
```
NightlyNexus: do we care about this case, anyway?

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -149,7 +149,7 @@
   };
 
   private final Listener listener;
-  private final RequestTransformer requestTransformer;
+  private final List<RequestTransformer> requestTransformers;
```
NightlyNexus: `@Nullable`

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -168,13 +168,14 @@
   boolean shutdown;
 
   Picasso(Context context, Dispatcher dispatcher, Cache cache, Listener listener,
-      RequestTransformer requestTransformer, List<RequestHandler> extraRequestHandlers, Stats stats,
-      Bitmap.Config defaultBitmapConfig, boolean indicatorsEnabled, boolean loggingEnabled) {
+      List<RequestTransformer> requestTransformers, List<RequestHandler> extraRequestHandlers,
```
NightlyNexus: `@Nullable` on the list param.

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -168,13 +168,14 @@
...
+      Stats stats, Bitmap.Config defaultBitmapConfig, boolean indicatorsEnabled,
+      boolean loggingEnabled) {
     this.context = context;
     this.dispatcher = dispatcher;
     this.cache = cache;
     this.listener = listener;
-    this.requestTransformer = requestTransformer;
+    this.requestTransformers = Collections.unmodifiableList(new ArrayList<>(requestTransformers));
```
NightlyNexus: refactoring bug?
`this.requestTransformers = requestTransformers != null ? Collections.unmodifiableList(new ArrayList<>(requestTransformers)) : null;`

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: depends on the transformer impl?

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: yep, good catch.

picasso/src/main/java/com/squareup/picasso/Picasso.java
NightlyNexus: i mean how likely is it that somebody accidentally adds the same RequestTransformer twice?

jrodbx: bugs happen :)

picasso/src/main/java/com/squareup/picasso/Picasso.java
NightlyNexus: sorry, just realized this is Picasso.Builder. was thinking about the Request.Builder.
let's do a non-null list. you're just creating a shared Picasso instance.

NightlyNexus: sorry about that.

NightlyNexus:
i think somebody said somewhere, but remind me again what the motivation is.

could a user write a single RequestTransformer, moving the loop to user code instead?
or, is this in preparation for more functionality in the chain?

picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -471,14 +473,23 @@ public void shutdown() {
...
-      throw new IllegalStateException("Request transformer "
-          + requestTransformer.getClass().getCanonicalName()
-          + " returned null for "
-          + request);
+    if (requestTransformers == null) {
+      return request;
     }
-    return transformed;
+
+    int numTransformers = requestTransformers.size();
```
JakeWharton: I typically put this into the loop
```
for (int i = 0, size = requestTransformers.size(); i < size; i++) {
```

jrodbx:
@NightlyNexus 
yep, Jake mentioned it here: https://github.com/square/picasso/issues/1800#issuecomment-371693216.  The idea is to mimic OkHttp's interceptor model and allow each transformer to potentially modify requests as it propagates down the chain.

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -678,7 +674,7 @@ void shutdown() {
     private ExecutorService service;
     private Cache cache;
     private Listener listener;
-    private RequestTransformer transformer;
+    private List<RequestTransformer> requestTransformers = new ArrayList<>();
```
NightlyNexus: nit: can be `final`

jrodbx: updated

Issue #1797, opened 2018-03-08
Request.networkPolicy is never assigned.


NightlyNexus:
This is broken. :(

NightlyNexus:
https://github.com/square/picasso/blob/6b18cb8599c83d288e59595fcc453db386cc85f1/picasso/src/main/java/com/squareup/picasso/NetworkRequestHandler.java#L51
no way that's reporting the right thing now.

jrodbx:
What's broken here?

NightlyNexus:
`throw new ResponseException(response.code(), request.networkPolicy)`
request.networkPolicy will always be 0. the field is never assigned.

jrodbx:
🤦‍♂️

PR #1802, opened 2018-03-09
Set network and memory policy on Request.

NightlyNexus:
Not sure how to test this.
https://github.com/square/picasso/issues/1800

Closes https://github.com/square/picasso/issues/1797 but it would be nice to have a test.

note that this lets you set a MemoryPolicy.NO_CACHE in a RequestTransformer which sometimes has an effect and sometimes doesn't.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Request.java
@@ -39,7 +39,8 @@ public final class Request {
   /** The time that the request was first submitted (in nanos). */
   long started;
   /** The {@link NetworkPolicy} to use for this request. */
-  int networkPolicy;
+  final int networkPolicy;
+  final int memoryPolicy;
 
   /**
    * The image URI.
@@ -98,7 +99,8 @@ public final class Request {
       int targetWidth, int targetHeight, boolean centerCrop, boolean centerInside,
       int centerCropGravity, boolean onlyScaleDown, float rotationDegrees,
       float rotationPivotX, float rotationPivotY, boolean hasRotationPivot,
-      boolean purgeable, Bitmap.Config config, Priority priority) {
+      boolean purgeable, Bitmap.Config config, Priority priority,
+      int networkPolicy, int memoryPolicy) {
     this.uri = uri;
     this.resourceId = resourceId;
     this.stableKey = stableKey;
@@ -120,6 +122,8 @@ public final class Request {
     this.purgeable = purgeable;
     this.config = config;
     this.priority = priority;
+    this.networkPolicy = networkPolicy;
+    this.memoryPolicy = memoryPolicy;
   }
 
   @Override public String toString() {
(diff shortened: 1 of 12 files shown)


jrodbx:
The more I think about this, the more I can't come up with a practical use case.  Maybe we leave this open or close/revisiting later when we have one?

NightlyNexus:
I think you're right. I'll take another look later tonight.

NightlyNexus:
Removed the Request.Builder APIs. Opinions on this as a fix for #1797? (can rename commits)

picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -480,6 +492,54 @@ public Builder transform(@NonNull List<? extends Transformation> transformations
...
+    Builder networkPolicy(@NonNull NetworkPolicy policy,
+        @NonNull NetworkPolicy... additional) {
+      if (policy == null) {
+        throw new IllegalArgumentException("Network policy cannot be null.");
+      }
+      this.networkPolicy |= policy.index;
+      if (additional == null) {
+        throw new IllegalArgumentException("Network policy cannot be null.");
+      }
+      if (additional.length > 0) {
```
swankjesse: nit: this if check is not necessary

picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -480,6 +492,54 @@ public Builder transform(@NonNull List<? extends Transformation> transformations
...
+    Builder memoryPolicy(@NonNull MemoryPolicy policy,
+        @NonNull MemoryPolicy... additional) {
+      if (policy == null) {
+        throw new IllegalArgumentException("Memory policy cannot be null.");
+      }
+      this.memoryPolicy |= policy.index;
+      if (additional == null) {
+        throw new IllegalArgumentException("Memory policy cannot be null.");
+      }
+      if (additional.length > 0) {
```
swankjesse: nit: this if check is not necessary

picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -480,6 +492,54 @@ public Builder transform(@NonNull List<? extends Transformation> transformations
...
     }
 
+    /**
+     * Specifies the {@link MemoryPolicy} to use for this request. You may specify additional policy
+     * options using the varargs parameter.
+     */
+    Builder memoryPolicy(@NonNull MemoryPolicy policy,
+        @NonNull MemoryPolicy... additional) {
+      if (policy == null) {
+        throw new IllegalArgumentException("Memory policy cannot be null.");
```
swankjesse: "policy == null" ?

picasso/src/main/java/com/squareup/picasso3/Request.java
swankjesse: though this is not new code!

picasso/src/main/java/com/squareup/picasso3/Request.java
JakeWharton: Also NPE not IAE

NightlyNexus: Yeah, I can fix these. Just existing code I pasted.

jrodbx:
Closed in https://github.com/square/picasso/pull/1856.

NightlyNexus:
nice, thanks!

PR #1804, opened 2018-03-09
Remove OkHttp3Downloader.

### LINKED ISSUE #1803: Remove the Downloader completely.
I don't think it need to exist. My gut says the tests are going to go haywire, though, and create a big diff.
still gonna try it.

NightlyNexus:
OkHtt3Downloader held the call factory and the cache but didn't do anything with them together. So, I put the cache and call factory directly into the places that use them.
Closes #1803.


picasso/src/main/java/com/squareup/picasso/Picasso.java
```diff
@@ -821,14 +831,16 @@ public Builder loggingEnabled(boolean enabled) {
...
-        okhttp3.Cache cache = new okhttp3.Cache(cacheDir, maxSize);
-        OkHttpClient client = new OkHttpClient.Builder()
-            .cache(cache)
+        unsharedCache = new okhttp3.Cache(cacheDir, maxSize);
+        callFactory = new OkHttpClient.Builder()
+            .cache(unsharedCache)
             .build();
-        downloader = new OkHttp3Downloader(client, cache, false);
+      } else {
+        unsharedCache = null;
```
JakeWharton: Initialize to `null` and then no `else` is needed.

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -1162,4 +1164,10 @@
       return new Result(bitmap, MEMORY);
     }
   }
+
+  private static final Call.Factory UNUSED_CALL_FACTORY = new Call.Factory() {
```
JakeWharton: Delete and use the one in `TestUtils`

jrodbx: TestUtils only has a fake `OkHttp3Downloader`, but we could pull out the `Call.Factory` and use that here.  👍

JakeWharton: This PR does that. It defines it in both places.

picasso/src/main/java/com/squareup/picasso/Picasso.java
jrodbx: I'd just inline `unsharedCache` into the builder.

picasso/src/test/java/com/squareup/picasso/TestUtils.java
```diff
@@ -274,11 +274,11 @@ static Bitmap makeBitmap(int width, int height) {
     return Bitmap.createBitmap(width, height, ALPHA_8);
   }
 
-  static final OkHttp3Downloader UNUSED_DOWNLOADER = new OkHttp3Downloader(new Call.Factory() {
-    @Override public Call newCall(okhttp3.Request request) {
+  static final Call.Factory UNUSED_CALL_FACTORY = new Call.Factory() {
+    @Override public Call newCall(@NonNull okhttp3.Request request) {
```
jrodbx: nit: remove @NonNull, let's document @Nullable with everything else defaulting to @NonNull.

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
jrodbx: Ah, missed that.  👍

picasso/src/test/java/com/squareup/picasso/BitmapHunterTest.java
```diff
@@ -83,14 +84,13 @@
 import static org.robolectric.Shadows.shadowOf;
 
 @RunWith(RobolectricGradleTestRunner.class)
-public class BitmapHunterTest {
+public final class BitmapHunterTest {
```
jrodbx: final test class?

jrodbx:
Merged in 6bd6ab8

Issue #1791, opened 2018-03-07
RequestCreator.get should cache the Bitmap.


NightlyNexus:
Unless somebody disagrees, I can update this.

JakeWharton:
It should. It represents a fundamental problem with the architecture that
needs fixed.

jrodbx:
Fixed in #1805.

PR #1805, opened 2018-03-09
Cache the result of RequestCreator.get.

NightlyNexus:
I wanted to test this like:

```java
@Test public void getCaches() throws Exception {
  Picasso picasso = new Picasso.Builder(context).build();
  picasso.load(URI_1).get();
  final AtomicBoolean success = new AtomicBoolean();
  picasso.load(URI_1).fetch(new Callback() {
    @Override public void onSuccess() {
      success.set(true);
    }

    @Override public void onError(Exception e) {

    }
  });
  assertThat(success.get()).isTrue();
}
```
but `get` can't be called on the main thread and onSuccess or onFailure won't ever be called, anyway.

Any ideas?

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/RequestCreator.java
+++ b/picasso/src/main/java/com/squareup/picasso3/RequestCreator.java
@@ -37,6 +37,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.squareup.picasso3.BitmapHunter.forRequest;
 import static com.squareup.picasso3.MemoryPolicy.shouldReadFromMemoryCache;
+import static com.squareup.picasso3.MemoryPolicy.shouldWriteToMemoryCache;
 import static com.squareup.picasso3.Picasso.LoadedFrom.MEMORY;
 import static com.squareup.picasso3.Picasso.Priority;
 import static com.squareup.picasso3.PicassoDrawable.setBitmap;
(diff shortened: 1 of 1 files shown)


NightlyNexus:
Closes #1791 (unless I missed something bigger).

picasso/src/main/java/com/squareup/picasso/RequestCreator.java
```diff
@@ -426,7 +426,10 @@ public Bitmap get() throws IOException {
     String key = createKey(finalData, new StringBuilder());
 
     Action action = new GetAction(picasso, finalData, memoryPolicy, networkPolicy, tag, key);
-    return forRequest(picasso, picasso.dispatcher, picasso.cache, picasso.stats, action).hunt();
+    Bitmap result =
+        forRequest(picasso, picasso.dispatcher, picasso.cache, picasso.stats, action).hunt();
+    picasso.insertIntoMemoryCache(key, result);
```
jrodbx: this doesn't respect the caching strategy.  maybe revert `insertIntoMemoryCache` and instead do:
```
   if (shouldWriteToMemoryCache(action.getMemoryPolicy())) {
     picasso.cache.set(key, result);
   }
```
WDYT?

NightlyNexus: yes, sounds good.
also, i forgot to update the doc on `get()`.

jrodbx:
Merged in https://github.com/square/picasso/commit/9323a108d5a25acbd8f68b70208363358389c143.
Closes #1791.

PR #1811, opened 2018-03-09
Allow placeholder resIds in remote views

jrodbx:
Currently, adding the sample widget crashes due to this superfluous check.


JakeWharton:
Shouldn't the sample just be changed to not specify a placeholder then?

jrodbx:
It could, but upon testing, it actually works since normal/error/placeholder resIds all map to `RemoteViews.setImageViewResource(int viewId, int srcId)`.

JakeWharton:
Hmm remove the whole `if` then?

JakeWharton:
Oh it only works for IDs

PR #1814, opened 2018-03-10
Eliminate executor thread scaling.

JakeWharton:
We're moving to using OkHttp's asynchronous calls and threads and so Picasso will only be responsible for Bitmap decoding. This means that the load is not correlated to the network anymore.


JakeWharton:
The `Dispatcher` is mutable so you can scale its pool size in the exact same way. And yes, this was an amazing oversimplification of something that's really complex to tune. I'm sure we could spend a month just figuring out all the knobs. But with this change, it becomes OkHttp's problem to provide the nobs, not ours!

PR #1816, opened 2018-03-10
Passing null should throw NPE.

JakeWharton:
This is consistent with what we do on all other projects.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Picasso.java
@@ -57,6 +57,7 @@ import static com.squareup.picasso3.Utils.VERB_ERRORED;
 import static com.squareup.picasso3.Utils.VERB_RESUMED;
 import static com.squareup.picasso3.Utils.calculateDiskCacheSize;
 import static com.squareup.picasso3.Utils.checkMain;
+import static com.squareup.picasso3.Utils.checkNotNull;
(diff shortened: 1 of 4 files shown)


picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -343,7 +332,7 @@ public RequestCreator load(@Nullable String path) {
    * @see #load(String)
    * @see #load(int)
    */
-  public RequestCreator load(@NonNull File file) {
+  public RequestCreator load(@Nullable File file) {
```
NightlyNexus: whoops. 👍

picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -434,18 +435,14 @@ public Builder purgeable() {
 
     /** Decode the image using the specified config. */
     public Builder config(@NonNull Bitmap.Config config) {
-      if (config == null) {
-        throw new IllegalArgumentException("config == null");
-      }
+      checkNotNull(config, "config == null");
```
NightlyNexus: any concerns about costs of string allocation/gc during request creation during, say, a list flinging?

JakeWharton: It's in the constant pool so it'll only be a `ldc` bytecode. I'm happy to inline these in a follow-up though.

PR #1817, opened 2018-03-10
Add versions plugin

jrodbx:
Also, bump Truth


build.gradle
```diff
@@ -1,3 +1,5 @@
+apply plugin: 'com.github.ben-manes.versions'
```
JakeWharton: It really weird me out that you can put this above `buildscript`. Can we put it below so I can pretend like Gradle doesn't do insane things to make this work?

jrodbx: done.

PR #1818, opened 2018-03-10
Replace MarkableInputStream with a superior Okio version.

JakeWharton:
This new `InputStream` uses the `Buffer` of the supplied `BufferedSource` to ensure it has buffered the necessary bytes to fulfill the reads of the consumer (via copies). This allows the original source to still be consumed normally by virtue of the fact that we never read directly from it.

There's a host of tests that we need for this but I'm holding off on writing them because we might actually upstream a mechanism similar to this into Okio.


JakeWharton:
Sample works, too.

picasso/src/test/java/com/squareup/picasso3/SourceBufferingInputStreamTest.java
```diff
@@ -0,0 +1,57 @@
...
+    // Read a 5 byte header from the wrapping input stream.
+    byte[] out = new byte[5];
+    assertEquals(5, stream.read(out));
+    assertEquals("Hello", new String(out, "UTF-8"));
+
+    // Ensure we've only requested what we needed from the upstream.
+    assertEquals(5, source.buffer().size());
+
+    // Ensure we can read the entirety of the contents from the original source.
+    assertEquals("Hello, world!", source.readUtf8());
```
jrodbx: nice

picasso/src/main/java/com/squareup/picasso3/SourceBufferingInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+  @Override public void reset() throws IOException {
+    if (mark == -1) {
+      throw new IOException("No mark or mark expired");
+    }
+    position = mark;
+    mark = -1;
+  }
+
+  @Override public int available() {
+    return (int) Math.max(buffer.size() - position, Integer.MAX_VALUE);
```
NightlyNexus: Is this right? I'm fairly certain this is supposed to `Math.min`.

JakeWharton: Yes. No one will ever hit it. I almost just did a straight cast.

NightlyNexus: for posterity: #1823

PR #1824, opened 2018-03-11
Add Error Prone and fix or suppress warnings.

NightlyNexus:


build.gradle
```diff
@@ -73,4 +75,12 @@ subprojects {
   afterEvaluate {
     tasks.findByName('check').dependsOn('checkstyle')
   }
-}
\ No newline at end of file
+
+  if (isCi) {
```
JakeWharton: Let's just always enable this. I only do conditional on projects that might be affected by the slower compiles.

build.gradle
```diff
@@ -27,21 +28,22 @@ buildscript {
...
+    classpath 'net.ltgt.gradle:gradle-errorprone-plugin:0.0.13'
   }
 }
 
 apply plugin: 'com.github.ben-manes.versions'
 
 subprojects {
   repositories {
-    jcenter()
+    mavenCentral()
```
jrodbx: does the plugin require this?  I thought plugins only require buildscript dependencies.

jrodbx: if not, let's remove.  else move under `google()`.

build.gradle
```diff
@@ -73,4 +75,10 @@ subprojects {
...
     tasks.findByName('check').dependsOn('checkstyle')
   }
-}
\ No newline at end of file
+
+  apply plugin: 'net.ltgt.errorprone'
+
+  dependencies {
+    errorprone deps.errorProne
+  }
```
jrodbx: I'm guessing this is related to the buildscript only comment.

picasso-provider/src/main/java/com/squareup/picasso3/provider/PicassoProvider.java
```diff
@@ -22,7 +22,7 @@
  */
 public final class PicassoProvider {
   @SuppressLint("StaticFieldLeak")
-  private static Picasso instance;
+  private static volatile Picasso instance;
```
jrodbx: hmm. would we ever want this to be accessed off the main thread?

picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
```diff
@@ -32,7 +32,7 @@
 
   private final Context context;
   private final Object lock = new Object();
-  private AssetManager assetManager;
+  private volatile AssetManager assetManager;
```
jrodbx: ditto

picasso/src/main/java/com/squareup/picasso3/FetchAction.java
```diff
@@ -19,6 +19,7 @@
 
 class FetchAction extends Action<Object> {
 
+  @SuppressWarnings("HidingField")
```
jrodbx: what's this do?

picasso/src/main/java/com/squareup/picasso3/Utils.java
```diff
@@ -336,8 +337,7 @@ static void flushStackLocalLeaks(Looper looper) {
   }
 
   static class PicassoThreadFactory implements ThreadFactory {
-    @SuppressWarnings("NullableProblems")
-    public Thread newThread(Runnable r) {
+    @Override public Thread newThread(@NonNull Runnable r) {
```
jrodbx: we've been removing NonNulls in favor of defaults, and only marking Nullable parameters.

picasso/src/test/java/com/squareup/picasso3/RequestCreatorTest.java
```diff
@@ -180,7 +180,7 @@ public void intoTargetWithFitThrows() {
     }
   }
 
-  public void intoTargetNoPlaceholderCallsWithNull() {
+  @Test public void intoTargetNoPlaceholderCallsWithNull() {
```
jrodbx: haha

picasso-provider/src/main/java/com/squareup/picasso3/provider/PicassoProvider.java
NightlyNexus: Probably not. Just matches what existed before.

picasso/src/main/java/com/squareup/picasso3/FetchAction.java
NightlyNexus: super.target is a different thing.

picasso/src/main/java/com/squareup/picasso3/Utils.java
NightlyNexus: IntelliJ gives a warning here cuz the supertype parameter is marked/known non-null.

build.gradle
NightlyNexus: Unrelated change, yeah.

NightlyNexus: okay, looking on my computer now...
this change isn't related to the plugin. i just figured we prefer mc over jcenter.
we need one of them for the dependencies like OkHttp, etc.
`google()` is below.

what are you asking?

build.gradle
NightlyNexus: i don't follow. what comment?

picasso-provider/src/main/java/com/squareup/picasso3/provider/PicassoProvider.java
NightlyNexus: to match #882

build.gradle
jrodbx: ah, never mind, read the errorprone README: https://github.com/tbroyer/gradle-errorprone-plugin#usage

picasso/src/main/java/com/squareup/picasso3/FetchAction.java
jrodbx: I renamed this and the other field to resolve the hidden/scoping issue.

jrodbx:
Merged in 3a883f052eca677a1c047c0d9e2b43a3bdaa1e4e

build.gradle
jrodbx: understood.  when I originally reviewed this (on mobile), I thought this diff related to the subject line of this PR ("errorprone").

PR #1827, opened 2018-03-12
Add JSR 305 nullability annotations.

jrodbx:


picasso-sample/src/main/java/com/example/picasso/SampleContactsActivity.java
```diff
@@ -84,7 +85,7 @@ private void loadContacts() {
           null, //
           ContactsQuery.SORT_ORDER);
     }
-    return null;
+    return new CursorLoader(this);
```
JakeWharton: throw instead?

jrodbx:
Ah, hadn't realized that re: worsening the Kotlin experience.  Is there a better way?

NightlyNexus:
While it'd be nice to remove the pollution of non-null annotations, I feel weird about mixing two nullability annotations.

I'd be in favor of either leaving the non-null annotations (I think that's what SQLDelight does) or implicitly saying that all types are non-null unless otherwise stated.

JakeWharton:
Yeah I think I would prefer to just be explicit unless a successor to JSR
305 emerges and is actually made a standard. We can add Uber's NullAway to
force our hand.

jrodbx:
Ok, I'll close.  For context, I was aiming to mimic what Retrofit did here: https://github.com/square/retrofit/commit/3c963a4a470194961e98952b2b51b8b29901d9e5.

PR #1838, opened 2018-03-21
Convert RequestHandler loading to async methods

jrodbx:
Need to cleanup BitmapHunter.hunt and its associated test, but wanted to get some early eyes on this.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
+++ b/picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
@@ -18,7 +18,6 @@ package com.squareup.picasso3;
 import android.content.Context;
 import android.content.res.AssetManager;
 import android.net.Uri;
-import java.io.IOException;
 import okio.Okio;
 import okio.Source;
 
@@ -26,7 +25,7 @@ import static android.content.ContentResolver.SCHEME_FILE;
 import static com.squareup.picasso3.Picasso.LoadedFrom.DISK;
 
 class AssetRequestHandler extends RequestHandler {
-  protected static final String ANDROID_ASSET = "android_asset";
+  private static final String ANDROID_ASSET = "android_asset";
   private static final int ASSET_PREFIX_LENGTH =
       (SCHEME_FILE + ":///" + ANDROID_ASSET + "/").length();
 
@@ -44,7 +43,7 @@ class AssetRequestHandler extends RequestHandler {
         && !uri.getPathSegments().isEmpty() && ANDROID_ASSET.equals(uri.getPathSegments().get(0)));
   }
 
-  @Override public Result load(Request request, int networkPolicy) throws IOException {
+  @Override public void load(Request request, int networkPolicy, Callback callback) {
     if (assetManager == null) {
       synchronized (lock) {
         if (assetManager == null) {
@@ -52,8 +51,16 @@ class AssetRequestHandler extends RequestHandler {
         }
       }
     }
-    Source source = Okio.source(assetManager.open(getFilePath(request)));
-    return new Result(source, DISK);
+    boolean signaledCallback = false;
+    try {
+      Source source = Okio.source(assetManager.open(getFilePath(request)));
+      signaledCallback = true;
+      callback.onSuccess(new Result(source, DISK));
+    } catch (Exception e) {
+      if (!signaledCallback) {
+        callback.onError(e);
+      }
+    }
   }
 
   static String getFilePath(Request request) {
--- a/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
@@ -20,15 +20,20 @@ import android.graphics.BitmapFactory;
 import android.graphics.Matrix;
 import android.net.NetworkInfo;
 import android.os.Build;
+import android.support.annotation.Nullable;
 import android.view.Gravity;
+import com.squareup.picasso3.RequestHandler.Result;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InterruptedIOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import okio.BufferedSource;
 import okio.Okio;
 import okio.Source;
@@ -74,8 +79,8 @@ class BitmapHunter implements Runnable {
       return true;
     }
 
-    @Override public Result load(Request request, int networkPolicy) throws IOException {
-      throw new IllegalStateException("Unrecognized type of request: " + request);
+    @Override public void load(Request request, int networkPolicy, Callback callback) {
+      callback.onError(new IllegalStateException("Unrecognized type of request: " + request));
     }
   };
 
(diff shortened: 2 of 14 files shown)


picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -209,52 +212,70 @@ Bitmap hunt() throws IOException {
...
-          }
-        }
-      }
+    Bitmap bitmap = null;
+    CountDownLatch latch = new CountDownLatch(1);
+    try {
+      latch.await();
+    }
+    catch (InterruptedException ie) {
+      throw new IOException(ie);
```
swankjesse: nit: use InterruptedIOException

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -209,52 +212,70 @@ Bitmap hunt() throws IOException {
...
-            //noinspection ConstantConditions If bitmap is null then source is guranteed non-null.
-            source.close();
-          } catch (IOException ignored) {
-          }
-        }
-      }
+    Bitmap bitmap = null;
+    CountDownLatch latch = new CountDownLatch(1);
+    try {
+      latch.await();
```
swankjesse: countDown not called?

swankjesse: and the local bitmap is shadowed by the one in the callback . . . I think you want an AtomicReference or something

picasso/src/main/java/com/squareup/picasso3/MediaStoreRequestHandler.java
```diff
@@ -53,47 +52,51 @@
...
+          // instead which is the largest thumbnail size can be fetched from MediaStore.
+          int kind = (picassoKind == FULL) ? Video.Thumbnails.MINI_KIND : picassoKind.androidKind;
+          bitmap = Video.Thumbnails.getThumbnail(contentResolver, id, kind, options);
+        } else {
+          bitmap =
+              Images.Thumbnails.getThumbnail(contentResolver, id, picassoKind.androidKind, options);
+        }
+
+        if (bitmap != null) {
+          callback.onSuccess(new Result(bitmap, null, DISK, exifOrientation));
```
swankjesse: needs a return after this line

picasso/src/main/java/com/squareup/picasso3/ContactsPhotoRequestHandler.java
```diff
@@ -65,12 +65,17 @@
...
+  @Override public void load(Request request, int networkPolicy, Callback callback) {
+    try {
+      InputStream is = getInputStream(request);
+      if (is == null) {
+        callback.onError(new IOException("no contact found"));
+      } else {
+        callback.onSuccess(new Result(Okio.source(is), DISK));
+      }
+    } catch (Exception e) {
+      callback.onError(e);
```
swankjesse: if `callback.onError()` throws, this calls `callback.onError()`. We need to do something separate to call the callback only 1x.

We have a `signalledCallback` thing in OkHttp when we have to solve this
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/RealCall.java#L145

picasso/src/main/java/com/squareup/picasso3/FileRequestHandler.java
```diff
@@ -37,9 +37,13 @@
...
 
-  @Override public Result load(Request request, int networkPolicy) throws IOException {
-    Source source = Okio.source(getInputStream(request));
-    return new Result(null, source, DISK, getFileExifRotation(request.uri));
+  @Override public void load(Request request, int networkPolicy, Callback callback) {
+    try {
+      Source source = Okio.source(getInputStream(request));
+      callback.onSuccess(new Result(null, source, DISK, getFileExifRotation(request.uri)));
+    } catch (Exception e) {
+      callback.onError(e);
```
swankjesse: same. Could be called back 2x

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java
```diff
@@ -42,30 +42,33 @@
...
-
-    // Cache response is only null when the response comes fully from the network. Both completely
-    // cached and conditionally cached responses will have a non-null cache response.
-    Picasso.LoadedFrom loadedFrom = response.cacheResponse() == null ? NETWORK : DISK;
+  @Override public void load(Request request, int networkPolicy, Callback callback) {
+    try {
+      okhttp3.Request callRequest = createRequest(request, networkPolicy);
+      Response response = callFactory.newCall(callRequest).execute();
+      if (!response.isSuccessful()) {
+        callback.onError(new ResponseException(response.code(), request.networkPolicy));
```
swankjesse: needs a return

swankjesse: needs a body.close() ?

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java
```diff
@@ -42,30 +42,33 @@
...
+      // Cache response is only null when the response comes fully from the network. Both completely
+      // cached and conditionally cached responses will have a non-null cache response.
+      Picasso.LoadedFrom loadedFrom = response.cacheResponse() == null ? NETWORK : DISK;
+
+      // Sometimes response content length is zero when requests are being replayed. Haven't found
+      // root cause to this but retrying the request seems safe to do so.
+      ResponseBody body = response.body();
+      if (loadedFrom == DISK && body.contentLength() == 0) {
+        body.close();
+        callback.onError(
```
swankjesse: needs a return

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java
```diff
@@ -42,30 +42,33 @@
...
+            new ContentLengthException("Received response with 0 content-length header."));
+      }
+      if (loadedFrom == NETWORK && body.contentLength() > 0) {
+        stats.dispatchDownloadFinished(body.contentLength());
+      }
+      callback.onSuccess(new Result(body.source(), loadedFrom));
+    } catch (IOException e) {
+      callback.onError(e);
```
swankjesse: could be signaled multiple times

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
```diff
@@ -119,11 +130,18 @@
...
     Action action = TestUtils.mockAction(URI_KEY_1, URI_1);
-    networkHandler.load(action.getRequest(), 0);
-    verify(stats).dispatchDownloadFinished(10);
+    networkHandler.load(action.getRequest(), 0, new Callback() {
+      @Override public void onSuccess(Result result) {
+        verify(stats).dispatchDownloadFinished(10);
+      }
+
+      @Override public void onError(Throwable t) {
+        fail(t.getMessage());
```
swankjesse: tricky question – does this route to the right thread?

picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
```diff
@@ -44,16 +43,20 @@
...
         }
       }
     }
-    Source source = Okio.source(assetManager.open(getFilePath(request)));
-    return new Result(source, DISK);
+    try {
+      Source source = Okio.source(assetManager.open(getFilePath(request)));
+      callback.onSuccess(new Result(source, DISK));
+    } catch (Exception e) {
+      callback.onError(e);
```
NightlyNexus: callback could be called twice. did you mean to catch only the IOExceptions from line 55?

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
NightlyNexus: yeah, think this is on the wrong line.
looks like it's meant to countDown and set an AtomicReference in the callback below and then await to return resultReference.get().

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
NightlyNexus: nit: move catch onto same line as the ending try bracket.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -209,52 +212,70 @@ Bitmap hunt() throws IOException {
...
-          bitmap = decodeStream(source, data);
-        } finally {
-          try {
-            //noinspection ConstantConditions If bitmap is null then source is guranteed non-null.
-            source.close();
-          } catch (IOException ignored) {
-          }
-        }
-      }
+    Bitmap bitmap = null;
```
NightlyNexus: replace with AtomicReference, yeah?

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -209,52 +212,70 @@ Bitmap hunt() throws IOException {
...
+            if (bitmap != null) {
+              stats.dispatchBitmapTransformed(bitmap);
+            }
+          }
         }
       }
-    }
+
+      @Override public void onError(Throwable t) {
+        throw new IOException(t);
```
NightlyNexus: what's going on here?
this looks like it's meant to throw in BitmapHunter.load. is this supposed to be in an `AtomicReference<Exception>`?

picasso/src/main/java/com/squareup/picasso3/Callback.java
```diff
@@ -15,17 +15,19 @@
  */
 package com.squareup.picasso3;
 
+import com.squareup.picasso3.RequestHandler.Result;
+
 public interface Callback {
-  void onSuccess();
+  void onSuccess(Result result);
```
NightlyNexus: nullable Result?

picasso/src/main/java/com/squareup/picasso3/ContactsPhotoRequestHandler.java
NightlyNexus: i'm not sure the internal implementation of this callback is supposed to throw?...

picasso/src/main/java/com/squareup/picasso3/ContentStreamRequestHandler.java
```diff
@@ -37,9 +36,13 @@
...
 
-  @Override public Result load(Request request, int networkPolicy) throws IOException {
-    Source source = Okio.source(getInputStream(request));
-    return new Result(source, DISK);
+  @Override public void load(Request request, int networkPolicy, Callback callback) {
+    try {
+      Source source = Okio.source(getInputStream(request));
+      callback.onSuccess(new Result(source, DISK));
+    } catch (Exception e) {
+      callback.onError(e);
```
NightlyNexus: same thing here. double callback call.

picasso/src/main/java/com/squareup/picasso3/MediaStoreRequestHandler.java
```diff
@@ -53,47 +52,51 @@
...
+        }
       }
 
-      if (bitmap != null) {
-        return new Result(bitmap, null, DISK, exifOrientation);
-      }
+      Source source = Okio.source(getInputStream(request));
+      callback.onSuccess(new Result(null, source, DISK, exifOrientation));
+    } catch (Exception e) {
+      callback.onError(e);
```
NightlyNexus: ditto earlier comments.

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
NightlyNexus: throw an assertion error and the countdownlatch.countDown won't be called, anyway.

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
```diff
@@ -119,11 +130,18 @@
...
 
-  @Test public void noCacheAndKnownContentLengthDispatchToStats() throws Exception {
+  @Test public void noCacheAndKnownContentLengthDispatchToStats() {
     responses.add(responseOf(ResponseBody.create(null, new byte[10])));
     Action action = TestUtils.mockAction(URI_KEY_1, URI_1);
-    networkHandler.load(action.getRequest(), 0);
-    verify(stats).dispatchDownloadFinished(10);
+    networkHandler.load(action.getRequest(), 0, new Callback() {
+      @Override public void onSuccess(Result result) {
+        verify(stats).dispatchDownloadFinished(10);
```
NightlyNexus: have a countdownlatch, set an atomic reference here, latch.countDown here, latch.await(10, SECONDS) on the test's main thread, and then verify the result in the atomic reference on the test's main thread.

otherwise, this test won't fail because the assertions are off the test's main thread.

picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
JakeWharton: I think we probably want to design the internal callback API that way so it works more like Rx.

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java
jrodbx: updated.  re:`close()`, I moved the `response.body()` call lower; `response.code()` doesn't consume the body.

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java
jrodbx: fixed.

picasso/src/main/java/com/squareup/picasso3/FileRequestHandler.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/ContactsPhotoRequestHandler.java
jrodbx: probably shouldn't, but added just in case.

picasso/src/main/java/com/squareup/picasso3/MediaStoreRequestHandler.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: yea, this was incomplete.  fixed now.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: done and done.

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java
jrodbx: fixed.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/Callback.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: good point.  Yes, it should throw in BitmapHunter.  Thoughts on wrapping in a `RuntimeException `which would get propagated back up to `BitmapHunter.load()`?

picasso/src/main/java/com/squareup/picasso3/AssetRequestHandler.java
jrodbx: @JakeWharton can tackle in a follow-up PR.

jrodbx:
@JakeWharton @swankjesse @NightlyNexus PTAAL.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: Updated to use an AtomicReference<Throwable>.

PR #1843, opened 2018-03-25
Push Picasso down into RequestHandler

jrodbx:
Trying to figure out why Mockito tests are failing...

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
+++ b/picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
@@ -35,8 +35,8 @@ import static com.squareup.picasso3.Utils.checkNotNull;
  * <p>
  * <h2>Usage</h2>
  * {@code RequestHandler} must be subclassed to be used. You will have to override two methods
- * ({@link #canHandleRequest(Request)} and {@link #load(Request, int, Callback)}) with your custom
- * logic to load images.
+ * ({@link #canHandleRequest(Request)} and {@link #load(Picasso, Request, int, Callback)}) with
+ * your custom logic to load images.
  * <p>
  * You should then register your {@link RequestHandler} using
  * {@link Picasso.Builder#addRequestHandler(RequestHandler)}
@@ -48,11 +48,11 @@ import static com.squareup.picasso3.Utils.checkNotNull;
  */
 public abstract class RequestHandler {
   /**
-   * {@link Result} represents the result of a {@link #load(Request, int, Callback)} call in a
-   * {@link RequestHandler}.
+   * {@link Result} represents the result of a {@link #load(Picasso, Request, int, Callback)} call
+   * in a {@link RequestHandler}.
    *
    * @see RequestHandler
-   * @see #load(Request, int, Callback)
+   * @see #load(Picasso, Request, int, Callback)
    */
   public static final class Result {
     private final Picasso.LoadedFrom loadedFrom;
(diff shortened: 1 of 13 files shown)


picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
```diff
@@ -107,11 +108,12 @@ int getExifOrientation() {
 
   /**
    * Loads an image for the given {@link Request}.
+   * @param picasso
```
swankjesse: nit: unnecessary

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
```diff
@@ -107,11 +108,12 @@ int getExifOrientation() {
...
    * @param request the data from which the image should be resolved.
    * @param networkPolicy the {@link NetworkPolicy} for this request.
    * @param callback
```
swankjesse: same

swankjesse:
Does `eq()` work if the argument to it is itself a mock? Hmm

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
jrodbx: updated.

jrodbx:
Good catch; however, changing `picasso` from a mock to a real impl doesn't resolve the issue.

runningcode:
Ok, no idea why this works, but it does
```
diff --git a/picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.java b/picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.java
index 6f99a67..90277f2 100644
--- a/picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.java
+++ b/picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.java
@@ -161,9 +161,10 @@ public final class BitmapHunterTest {
         new TestableBitmapHunter(picasso, dispatcher, cache, stats, action, bitmap);

     Bitmap result = hunter.hunt();
+    Request request = action.getRequest();
     assertThat(cache.missCount()).isEqualTo(1);
     verify(hunter.requestHandler).load(
-        eq(picasso), eq(action.getRequest()), eq(0), any(RequestHandler.Callback.class));
+        eq(picasso), eq(request), eq(0), any(RequestHandler.Callback.class));
     assertThat(result).isEqualTo(bitmap);
   }

@@ -179,8 +180,9 @@ public final class BitmapHunterTest {
         new TestableBitmapHunter(picasso, dispatcher, cache, stats, action, bitmap);

     Bitmap result = hunter.hunt();
+    Request request = action.getRequest();
     assertThat(cache.hitCount()).isEqualTo(1);
-    verify(hunter.requestHandler, never()).load(eq(picasso), eq(action.getRequest()), eq(0),
+    verify(hunter.requestHandler, never()).load(eq(picasso), eq(request), eq(0),
         any(RequestHandler.Callback.class));
     assertThat(result).isEqualTo(bitmap);
   }
```

jrodbx:
Ahh, I see.  Looks Mockito doesn't like it when you interact with a mock (here, by calling `action.getRequest()`) while in the middle of verifying a mock (here, `hunter.requestHandler`) using argument matchers.  The problem didn't exist before since `action.getRequest()` used to be the first argument and therefore was the first matcher encountered during verification.

Extracting the call to a variable moves the interaction with `action` to a point prior to the verification of `hunter.requestHandler.load(...)`.  Moving the newly-added `picasso` parameter away from the first position (i.e., after `action.getRequest()`) also "fixes" the issue.

Thanks @runningcode!

PR #1847, opened 2018-03-28
Allow single app widget

### LINKED ISSUE #1846: Not able to update a single AppWidget
At the moment picasso only takes a whole int array of AppWidget ids which are updated in one batch. I would expect to be able to also give picasso a single widget id as int as well. Of course one can give picasso an array with only one int inside but it looks like kind of an oversight to me.

jrodbx:
Closes #1846


C0br4:
the javadoc is incorrect. not in all widget instances, only the one with the given id.

PR #1852, opened 2018-03-29
Delegate to ImageDecoder if >= P

jrodbx:


picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
```diff
@@ -209,18 +221,49 @@ static void calculateInSampleSize(int reqWidth, int reqHeight, int width, int he
...
+
+    return decodeStreamPreP(request, bufferedSource);
+  }
+
+  @TargetApi(28)
+  @SuppressLint("Override")
+  private static Bitmap decodeStreamP(final Request request, BufferedSource bufferedSource)
+      throws IOException {
+    ImageDecoder.Source imageSource =
+        ImageDecoder.createSource(ByteBuffer.wrap(bufferedSource.readByteArray()));
```
JakeWharton: Ugh

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
```diff
@@ -209,18 +221,49 @@ static void calculateInSampleSize(int reqWidth, int reqHeight, int width, int he
...
     BufferedSource bufferedSource = Okio.buffer(source);
 
+    if (Build.VERSION.SDK_INT >= 28) {
+      return decodeStreamP(request, bufferedSource);
+    }
+
+    return decodeStreamPreP(request, bufferedSource);
+  }
+
+  @TargetApi(28)
```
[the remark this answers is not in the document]
jrodbx: I went with that first, but hit and filed this Lint issue: https://issuetracker.google.com/issues/77246433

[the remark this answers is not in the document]
jrodbx: oh right, but then I added the SuppressLint to shut that up.  nvm.  will update.

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
jrodbx: I know...

jrodbx: Looks like there's a perfectly valid, yet hidden overload here: https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/graphics/java/android/graphics/ImageDecoder.java#900

🤔

NightlyNexus: why does ImageDecoder not have a streaming API?!

JakeWharton: It needs to read multiple times and have random access

PR #1853, opened 2018-03-29
Cleanup Request

jrodbx:


picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -219,7 +211,7 @@ public Builder newBuilder() {
     private float rotationPivotY;
     private boolean hasRotationPivot;
     private boolean purgeable;
-    private List<Transformation> transformations;
+    private List<Transformation> transformations = new ArrayList<>();
```
NightlyNexus: maybe it's alright to let this list be null to signify the empty list? for performance considerations?

NightlyNexus: since lots of Requests get created on a hot path.

JakeWharton: Yeah we probably need to retain this null pain. Especially since I don't think custom transformations are overwhelmingly common.

jrodbx: reverted null removal.

PR #1857, opened 2018-03-31
Apply exif orientation in ContentStreamRequestHandler

jrodbx:
Extracted from and to replace a custom request handler we use internally.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/ContentStreamRequestHandler.java
+++ b/picasso/src/main/java/com/squareup/picasso3/ContentStreamRequestHandler.java
@@ -18,12 +18,16 @@ package com.squareup.picasso3;
 import android.content.ContentResolver;
 import android.content.Context;
 import android.graphics.Bitmap;
+import android.support.media.ExifInterface;
 import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.InputStream;
 import okio.Okio;
 import okio.Source;
 
 import static android.content.ContentResolver.SCHEME_CONTENT;
+import static android.support.media.ExifInterface.ORIENTATION_NORMAL;
+import static android.support.media.ExifInterface.TAG_ORIENTATION;
 import static com.squareup.picasso3.Picasso.LoadedFrom.DISK;
 
 class ContentStreamRequestHandler extends RequestHandler {
@@ -46,8 +50,9 @@ class ContentStreamRequestHandler extends RequestHandler {
         return;
       }
       Bitmap bitmap = decodeStream(source, request);
+      int exifRotation = getExifOrientation(request);
       signaledCallback = true;
-      callback.onSuccess(new Result(bitmap, DISK));
+      callback.onSuccess(new Result(bitmap, DISK, exifRotation));
     } catch (Exception e) {
       if (!signaledCallback) {
         callback.onError(e);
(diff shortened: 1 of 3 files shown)


picasso/src/main/java/com/squareup/picasso3/ContentStreamRequestHandler.java
```diff
@@ -60,4 +65,11 @@ Source getSource(Request request) throws FileNotFoundException {
     InputStream inputStream = contentResolver.openInputStream(request.uri);
     return inputStream == null ? null : Okio.source(inputStream);
   }
+
+  protected int getExifOrientation(Request request) throws IOException {
+    ContentResolver contentResolver = context.getContentResolver();
+    InputStream inputStream = contentResolver.openInputStream(request.uri);
```
JakeWharton: needs try/finally/close

picasso/src/main/java/com/squareup/picasso3/ContentStreamRequestHandler.java
```diff
@@ -60,4 +65,11 @@ Source getSource(Request request) throws FileNotFoundException {
...
+    ExifInterface exifInterface = new ExifInterface(inputStream);
```
JakeWharton: Any idea how much this reads from the stream? I wonder if it would be beneficial to do something similar to what we do for network requests and use the buffering input stream so that we can read twice. Provided this doesn't consume the whole stream, that is...

jrodbx: Looks like it can consume the whole stream.  In ExifInterface#getJpegAttributes:
```            
 // EOI indicates the end of an image and in case of SOS, JPEG image stream starts and
 // the image data will terminate right after.
 if (marker == MARKER_EOI || marker == MARKER_SOS) {
     break;
 }
```

picasso/src/main/java/com/squareup/picasso3/ContentStreamRequestHandler.java
jrodbx: done.

blinkmacalahan:
To others that might realize `ContentStreamRequestHandler` lacks EXIF support in `v2.71828` and want a quick fix, here's a custom request handler. **Note: In order to use this you'll need to make a package com.squareup.picasso in order to access package private methods**

```
package com.squareup.picasso

import android.content.ContentResolver
import android.content.Context
import androidx.exifinterface.media.ExifInterface
import com.squareup.picasso.Picasso.LoadedFrom.DISK
import okio.Okio
import okio.Source
import java.io.FileNotFoundException
import java.io.IOException

class ExifContentStreamRequestHandler(val context: Context) :
    RequestHandler() {

    override fun canHandleRequest(data: Request): Boolean {
        return ContentResolver.SCHEME_CONTENT == data.uri.scheme
    }

    @Throws(IOException::class)
    override fun load(
        request: Request,
        networkPolicy: Int
    ): Result? {
        var result: Result? = null
        try {
            val source = getSource(request)
            val exifRotation = getExifOrientation(request)
            result = Result(null, source, DISK, exifRotation)
        } catch (e: FileNotFoundException) {
            e.printStackTrace()
        }
        return result
    }

    @Throws(FileNotFoundException::class)
    fun getSource(request: Request): Source? {
        val contentResolver = context.contentResolver
        val inputStream = contentResolver.openInputStream(request.uri)
        return if (inputStream == null) null else Okio.source(inputStream)
    }

    @Throws(IOException::class)
    private fun getExifOrientation(request: Request): Int {
        val contentResolver = context.contentResolver
        contentResolver.openInputStream(request.uri).use { inputStream ->
            val exifInterface =
                ExifInterface(inputStream!!)
            return exifInterface.getAttributeInt(
                ExifInterface.TAG_ORIENTATION,
                ExifInterface.ORIENTATION_NORMAL
            )
        }
    }
}
```

PR #1859, opened 2018-04-01
Move Request tagging to the actual Request

jrodbx:


picasso/src/test/java/com/squareup/picasso3/DispatcherTest.java
```diff
@@ -143,48 +143,23 @@
...
-
-    FetchAction fetchAction =
-        new FetchAction(mockPicasso(), new Request.Builder(URI_1).build(), pausedTag, callback);
-    dispatcher.performSubmit(fetchAction, false);
-    Exception e = new RuntimeException();
-    fetchAction.error(e);
-
-    verify(callback).onError(e);
-  }
-
```
jrodbx: These tests weren't useful, as they were checking whether fetchAction.complete/error called the appropriate callback method -- nothing to do with Dispatcher!

PR #1866, opened 2018-04-02
Introduce Target2 to wrap target-specific config

jrodbx:


picasso/src/main/java/com/squareup/picasso3/Target2.java
```diff
@@ -0,0 +1,40 @@
+package com.squareup.picasso3;
+
+import android.graphics.drawable.Drawable;
+import android.support.annotation.DrawableRes;
+import android.support.annotation.NonNull;
+
+public class Target2<T> {
```
jrodbx: Will convert to a Builder in a follow-up PR.

jrodbx:
trying to pull things out of RequestCreator that are Target-specific (as opposed to Request-specific), moving towards the Source-Target decoupling we discussed earlier.

picasso/src/main/java/com/squareup/picasso3/Target2.java
```diff
@@ -0,0 +1,40 @@
+package com.squareup.picasso3;
```
JakeWharton: License

PR #1878, opened 2018-04-10
Remove synthetic accessors.

NightlyNexus:
Ran the IJ inspection over the project.
Fwiw, I just don't use `private` in Java.

### DIFF
--- a/picasso-sample/src/main/java/com/example/picasso/PicassoSampleAdapter.java
+++ b/picasso-sample/src/main/java/com/example/picasso/PicassoSampleAdapter.java
@@ -66,7 +66,7 @@ final class PicassoSampleAdapter extends BaseAdapter {
     };
 
     private final Class<? extends Activity> activityClass;
-    private final String name;
+    final String name;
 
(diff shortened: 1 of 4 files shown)


NightlyNexus:
https://issuetracker.google.com/issues/37075533

JakeWharton:
And if my math calculations are correct, 11 comes directly after 10!

jrodbx:
Uh, I don't think we should weaken access for the sake of synthetic accessors.  Wouldn't they get stripped away anyway at build time?

JakeWharton:
Nope!

jrodbx:
Ah ok.  Also, read the issue that @NightlyNexus linked.  Makes sense!

NightlyNexus:
Evil plan disclosure:
My zealotry against private is actually for readability more than avoiding synthetic methods. Synthetic methods are merely an excuse to spread my mischief.

Code in a feature package or library knows its intention. Private is a noisy modifier that hasn't helped me in practice, past first learning Java. (I wish Kotlin defaulted to internal.)

A religious discussion for a very different thread, though, haha.

PR #1879, opened 2018-04-10
Remove single-transform lock.

JakeWharton:
Devices have massive amounts of RAM now in proportion to the image sizes they are decoding. Also this was speculatively added without empirical data to justify its presence.


jrodbx:
> Devices have massive amounts of RAM now

Does that mean lower-end devices would potentially suffer from this change?  Can't see how, but worth asking.

JakeWharton:
They'd be more at risk of OOM. If it becomes a problem in practice, we can do something, but if that's true you should probably adjust your thread pool size of the executor doing decoding.

PR #1881, opened 2018-04-11
Remove batching to the main thread.

JakeWharton:
Texture uploads to the GPU are a fixed cost and by batching we actually group that cost which might actually cause missed frame deadlines. Multiple posts in the same frame isn't really an issue either as layout is already batched. Also this was speculatively added without empirical data to justify its presence.

Can't get the tests doing what I want with ShadowLooper. Mockito doesn't work because `sendMessage` is final. Can probably use https://developer.android.com/reference/android/os/TestLooperManager.html if we move to instrumentation tests.


NightlyNexus:
Nice. Cherry-picked into our ancient fork on Picasso 2.
Massively helpful.

PR #1884, opened 2018-04-11
Enforce Kotlin interop check, add nullability annotations.

JakeWharton:


picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -397,7 +402,7 @@ public void invalidate(@NonNull File file) {
   }
 
   /** {@code true} if debug indicators should are displayed on images. */
-  @SuppressWarnings("UnusedDeclaration") public boolean areIndicatorsEnabled() {
+  @SuppressWarnings("UnusedDeclaration") public boolean getIndicatorsEnabled() {
```
jrodbx: revert?  boolean getter?

JakeWharton: This is a failure.

https://android.github.io/kotlin-guides/interop.html#property-prefixes

JakeWharton: and `isIndicatorsEnabled` is weird and broken english

picasso/src/main/java/com/squareup/picasso3/MediaStoreRequestHandler.java
```diff
@@ -45,13 +46,14 @@
...
 
-  @Override public boolean canHandleRequest(Request data) {
+  @Override public boolean canHandleRequest(@NonNull Request data) {
     final Uri uri = data.uri;
     return (SCHEME_CONTENT.equals(uri.getScheme())
             && MediaStore.AUTHORITY.equals(uri.getAuthority()));
   }
 
-  @Override public void load(Picasso picasso, Request request, Callback callback) {
+  @Override
```
jaredsburrows: `@Override` on same line?

jaredsburrows:
@JakeWharton @jrodbx Would you all accept PRs migrating those tests to Instrumentation tests similar to https://github.com/android/android-ktx?

picasso/src/main/java/com/squareup/picasso3/MediaStoreRequestHandler.java
JakeWharton: It doesn't fit

PR #1887, opened 2018-04-12
Make Transformation modify all Results, not just Bitmaps

jrodbx:
Moving towards the Grand Unified Pipeline™.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
+++ b/picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
@@ -62,15 +62,15 @@ public abstract class RequestHandler {
     private final Picasso.LoadedFrom loadedFrom;
     private final Bitmap bitmap;
     private final Drawable drawable;
-    private final int exifOrientation;
+    private final int exifRotation;
 
     public Result(@NonNull Bitmap bitmap, @NonNull Picasso.LoadedFrom loadedFrom) {
       this(checkNotNull(bitmap, "bitmap == null"), null, loadedFrom, 0);
     }
 
     public Result(@NonNull Bitmap bitmap, @NonNull Picasso.LoadedFrom loadedFrom,
-        int exifOrientation) {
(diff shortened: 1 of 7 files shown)


picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
```diff
@@ -125,7 +125,7 @@ public boolean hasDrawable() {
      * {@link #load(Picasso, Request, Callback)} call. This is only accessible to built-in
      * RequestHandlers.
      */
-    int getExifOrientation() {
+    public int getExifOrientation() {
```
JakeWharton: We might want to rename this EXIF _rotation_. Something to consider.

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
```diff
@@ -125,7 +125,7 @@ public boolean hasDrawable() {
      * {@link #load(Picasso, Request, Callback)} call. This is only accessible to built-in
```
JakeWharton: Delete this second sentence

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
jrodbx: I'll update this in the next PR, since renaming will touch files which have nothing to do with this PR.

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
jrodbx: Ditto.

jrodbx: Updated.

picasso/src/main/java/com/squareup/picasso3/RequestHandler.java
jrodbx: Updated.

PR #1888, opened 2018-04-12
Move matrix/exif transforms into its own Transformer

jrodbx:


picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -388,7 +371,9 @@ static BitmapHunter forRequest(Picasso picasso, Dispatcher dispatcher,
     return new BitmapHunter(picasso, dispatcher, cache, stats, action, ERRORING_HANDLER);
   }
 
-  static Result applyCustomTransformations(List<Transformation> transformations, Result result) {
+  static Result applyTransformations(Picasso picasso, Request data,
+      List<Transformation> transformations,
+      Result result) {
     if (!result.hasBitmap()) {
```
JakeWharton: How many people call this method? Can we move this outside and avoid allocations? Can be follow-up.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -199,24 +190,16 @@ Result hunt() throws IOException {
...
-          result = applyCustomTransformations(data.transformations, result);
-          if (picasso.loggingEnabled) {
-            log(OWNER_HUNTER, VERB_TRANSFORMED, data.logId(),
-                "from custom transformations");
-          }
-        }
+      if (data.needsMatrixTransform() || result.getExifOrientation() != 0) {
+        transformations.add(new MatrixTransformation(data));
+      }
+      if (data.hasCustomTransformations()) {
```
JakeWharton: Let's delete this method and normalize to `emptyList()` in the constructor (and back to null when calling `newBuilder`). This way we never need to check and can always call `addAll`.

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
+package com.squareup.picasso3;
```
JakeWharton: License

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
...
+
+import static android.support.media.ExifInterface.ORIENTATION_FLIP_HORIZONTAL;
+import static android.support.media.ExifInterface.ORIENTATION_FLIP_VERTICAL;
+import static android.support.media.ExifInterface.ORIENTATION_ROTATE_180;
+import static android.support.media.ExifInterface.ORIENTATION_ROTATE_270;
+import static android.support.media.ExifInterface.ORIENTATION_ROTATE_90;
+import static android.support.media.ExifInterface.ORIENTATION_TRANSPOSE;
+import static android.support.media.ExifInterface.ORIENTATION_TRANSVERSE;
+
+public class MatrixTransformation implements Transformation {
```
JakeWharton: final, not public

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
...
+  private final Request data;
+
+  public MatrixTransformation(Request data) {
```
JakeWharton: not public

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
...
+      return source;
+    }
+
+    Bitmap bitmap = transformResult(data, source.getBitmap(), source.getExifOrientation());
+
+    return new RequestHandler.Result(bitmap, source.getLoadedFrom(), source.getExifOrientation());
+  }
+
+  @Override public String key() {
+    return null;
```
JakeWharton: This works? We need to throw if someone does this. Can be a follow-up.

If we're going to make this behavior into a transformation then we should move the code which adds the data to they key into this class and return a proper key. Can also be a follow-up.

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  static Bitmap transformResult(Request data, Bitmap result, int exifOrientation) {
```
JakeWharton: Can be private now I think

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
...
+    return result;
+  }
+
+  private static boolean shouldResize(boolean onlyScaleDown, int inWidth, int inHeight,
+      int targetWidth, int targetHeight) {
+    return !onlyScaleDown || (targetWidth != 0 && inWidth > targetWidth)
+        || (targetHeight != 0 && inHeight > targetHeight);
+  }
+
+  static int getExifRotation(int orientation) {
```
JakeWharton: Private

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
```diff
@@ -0,0 +1,229 @@
...
+      case ORIENTATION_TRANSVERSE:
+        rotation = 270;
+        break;
+      default:
+        rotation = 0;
+    }
+    return rotation;
+  }
+
+  static int getExifTranslation(int orientation)  {
```
JakeWharton: private

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
JakeWharton: Allocations == new ArrayList

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
JakeWharton: Can be follow-up as well.

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: fixed

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: fixed

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: fixed

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: can't, cuz tests

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: fixed

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: fixed

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
JakeWharton: Can we add `@VisibleForTesting` then?

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: Updated.

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: updated

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
jrodbx: Making non-null string value for now.  Can you elaborate more on "then we should move the code which adds the data to they key into this class and return a proper key"?

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
jrodbx: removing call to result.hasBitmap from here, as it's already called outside in its sole call site: BitmapHunter#hunt.

jrodbx:
Changes made, with one comment seeking clarification.  LGTY?

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -206,24 +197,13 @@ Result hunt() throws IOException {
...
-
-        result = new Result(bitmap, result.getLoadedFrom(), exifOrientation);
-        if (data.hasCustomTransformations()) {
-          result = applyCustomTransformations(data.transformations, result);
-          if (picasso.loggingEnabled) {
-            log(OWNER_HUNTER, VERB_TRANSFORMED, data.logId(),
-                "from custom transformations");
-          }
-        }
+      List<Transformation> transformations = new ArrayList<>();
```
JakeWharton: can be pre-sized to `data.transformations.size() + 1`. one extra slot is better than incurring expansion

picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -116,7 +115,7 @@
     this.resourceId = builder.resourceId;
     this.stableKey = builder.stableKey;
     if (builder.transformations == null) {
-      this.transformations = null;
+      this.transformations = new ArrayList<>();
```
JakeWharton: `Collections.emptyList()`?

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.java
JakeWharton: Keys are supposed to contain the data which makes them unique. If you have a rotate transformation the key needs to contain the rotation information. Otherwise the cache will return a 180 rotated image for someone requesting 90.

In this case I was saying we should update they key to contain the necessary information that makes the transformation unique.

jrodbx: got it.  updated to "matrixTransformation()".

PR #1891, opened 2018-04-19
Updates SourceBufferingInputStream to conform to InputStream contract

rharter:


picasso/src/main/java/com/squareup/picasso3/SourceBufferingInputStream.java
```diff
@@ -33,33 +35,49 @@
...
   private long mark = -1;
+  private long markLimit = -1;
 
   SourceBufferingInputStream(BufferedSource source) {
     this.source = source;
     this.buffer = source.buffer();
   }
 
   private final Buffer temp = new Buffer();
+  // offset is the write offset in the dest array
```
JakeWharton: Can you move this comment inside the method so it doesn't get lost if shuffling occurs. Also please capitalize "offset" and add a period. [Comments should be sentences][1].

 [1]: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

rharter:
It should be noted that this comes from #1890 and makes `SourceBufferingInputStream` consumable by Android's BitmapFactory and the AndroidSVG library.

picasso/src/main/java/com/squareup/picasso3/SourceBufferingInputStream.java
```diff
@@ -33,33 +35,49 @@
...
   private int copyTo(byte[] sink, int offset, int byteCount) {
     // TODO replace this with https://github.com/square/okio/issues/362
-    buffer.copyTo(temp, offset, byteCount);
+    buffer.copyTo(temp, position, byteCount);
```
JakeWharton: I meant to circle back and fix this in Picasso, but I noticed this was a bug when trying to add `copyTo` to Okio. It becomes really annoying because we need two offsets in the API.

picasso/src/main/java/com/squareup/picasso3/SourceBufferingInputStream.java
```diff
@@ -33,33 +35,49 @@
...
+    if (position > markLimit) {
       mark = -1;
     }
     return value;
   }
 
   @Override public int read(@NonNull byte[] b, int off, int len) throws IOException {
-    source.require(position + len);
-    int copied = /*buffer.*/copyTo(b, off, len);
+    checkNotNull(b, "b is null");
```
JakeWharton: `"b == null"`

JakeWharton: Although I'm not sure we really need this explicit check. `b` is immediately dereferenced by the `if` statement below.

picasso/src/main/java/com/squareup/picasso3/SourceBufferingInputStream.java
```diff
@@ -33,33 +35,49 @@
   private final Buffer buffer;
   private long position;
   private long mark = -1;
```
JakeWharton: Should we call this `markPosition` then?

picasso/src/test/java/com/squareup/picasso3/SourceBufferingInputStreamTest.java
```diff
@@ -45,6 +47,61 @@
...
+    assertEquals(5, len);
+    assertArrayEquals(new byte[] {'H', 'e', 'l', 'l', 'o'}, bytes);
+
+    len = stream.read(bytes);
+    assertEquals(5, len);
+    assertArrayEquals(new byte[] {',', ' ', 'w', 'o', 'r'}, bytes);
+
+    len = stream.read(bytes);
+    assertEquals(3, len);
+    // last two bytes are out of range so untouched from previous run
```
JakeWharton: nit: capitalize, period

picasso/src/test/java/com/squareup/picasso3/SourceBufferingInputStreamTest.java
```diff
@@ -45,6 +47,61 @@
...
+
+    len = stream.read(bytes);
+    assertEquals(len, 4);
+    assertArrayEquals(new byte[] {'l', 'l', 'o', ','}, bytes);
+
+    try {
+      stream.reset();
+      fail("expected IOException on reset");
+    } catch (IOException e) {
+      // success
```
JakeWharton: we tend to name the catch variable "expected" so that we can omit this comment

JakeWharton: Or you could assert on the message to make sure we're getting the correct IOE
```java
assertEquals("No mark or whatever", e.getMessage());
```

picasso/src/main/java/com/squareup/picasso3/SourceBufferingInputStream.java
```diff
@@ -42,34 +43,48 @@
   private final Buffer temp = new Buffer();
   private int copyTo(byte[] sink, int offset, int byteCount) {
     // TODO replace this with https://github.com/square/okio/issues/362
-    buffer.copyTo(temp, offset, byteCount);
+    // `copyTo` treats offset as the read position, `read` treats offset as the write offset.
```
JakeWharton: Very nice

jrodbx:
Thanks!

Issue #1894, opened 2018-05-03
Is it required to keep thread leak cleaning tasks running all the time after Android 5.0?


bluesway:
In picasso.java, refer to the comments, the following code seems unnecessary after Android 5.0, would it be better to add a system version checking? (something like `android.os.Build.VERSION`)

It's kinda annoyed for the threads running all the time per second on modern Android devices, it's still impact the power consumption (minor, but still affected)

```java
@Override public void run() {
  Process.setThreadPriority(THREAD_PRIORITY_BACKGROUND);
  while (true) {
    try {
      // Prior to Android 5.0, even when there is no local variable, the result from
      // remove() & obtainMessage() is kept as a stack local variable.
      // We're forcing this reference to be cleared and replaced by looping every second
      // when there is nothing to do.
      // This behavior has been tested and reproduced with heap dumps.
      RequestWeakReference<?> remove =
          (RequestWeakReference<?>) referenceQueue.remove(THREAD_LEAK_CLEANING_MS);
      Message message = handler.obtainMessage();
      if (remove != null) {
        message.what = REQUEST_GCED;
        message.obj = remove.action;
        handler.sendMessage(message);
      } else {
        message.recycle();
      }    
    } catch (InterruptedException e) {
      break;
    } catch (final Exception e) {
      handler.post(new Runnable() {
        @Override public void run() {
          throw new RuntimeException(e);
        }
      });
      break;
    }
  }
}
```

JakeWharton:
Yeah it's pretty ridiculous to do it if we're above that API level, especially since the majority has massively shifted above that line.

Want to send a PR to no-op it on newer devices?

PR #1895, opened 2018-05-03
Only run CleanupThread under Lollipop (#1894)

bluesway:
It's better to avoid redundant threads running after Android Lollipop (5.0)


JakeWharton:
Looking at this now, it's not the right solution. On APIs newer than 21 we should be using the no-argument `remove()` method which will park the thread until a reference comes along.

bluesway:
seems this is not the code which triggered Dispatch and Stats threads per second, even stop entering the while-loop, the periodical threads were still alive.

JakeWharton:
CleanupThread is dead! Thanks for your PR though.

PR #1901, opened 2018-05-21
fix when get bitmap from memory cache but the bitmap is recycled

earthgee:
in my project i use picasso to load image,and i use the api Bitmap.recycle() somewhere.but i don't notice this bitmap is also used in memory cache.so next time i took same bitmap from memory cache to use it,but it is recycled and cause some crash.
  so i fix it when get it from memory cache and check if the bitmap is recycled.then this "error" bitmap will be delete from cache in the future.


JakeWharton:
Why don't you just remove the calls to recycle() instead since they're undermining the cache and your application clearly would benefit from it as the instance was requested again.

jrodbx:
Closing.  As mentioned, explicitly using `recycle()` will interfere with Picasso's caching internals.

PR #1908, opened 2018-05-31
Empty string & 'null' check for image path in load()

RahulSDeshpande:
* This update contains the empty string & ```null``` check for image path provided as a parameter to the ```load()``` method.
* ```TextUtils.isEmpty(stringVar)``` will check the whether the provided string object is ```null``` or is of length '0' (zero).
* This will help us avoid throwing of ```IllegalArgumentException``` in case of image path is empty.


RahulSDeshpande:
Please ignore this PR as the CI build failed due to some trailing white space ! :P
Raising a PR separately by removing white space.

Issue #899, opened 2015-02-19
Null validation and Empty validation in Picasso Image Path


neodroidpune:
It wired to observe that Picasso has not handled null and Empty url 
1) If null is passed in Picasso.load() it never loads error Image which is required
2) If blank image passed in   Picasso.load() app gets crashed gives Path must not be empty Exception happens 
Developer need to add validation for both the cases. 

Thanks in advance 
Looking forward your reply

mattprecious:
Passing `null` is not an error, so Picasso will either clear the `ImageView` or load your placeholder if you specify one. This is a required and expected behaviour for view recycling.

An empty path is not valid, so Picasso throws an exception. What validation do you want here?

crispi7:
Please, just make your code better. Why are you passing an empty url?????

neodroidpune:
Urls are received from webservices , So if for certain webservice , image url is not there , it might return null or nothing. I think for these cases it should give error Image instead of throwing exception.

JakeWharton:
You can filter empty URLs to null before passing them to Picasso.

neodroidpune:
ya sure I can I just curious that why Picasso doesnt have such validation ,
Thanks for support.

JakeWharton:
Because while null is a very explicit thing, an empty URL has no meaning
and is always almost a sign of an error or bad data.

SriKrishnaCovalense:
Whatever is the bad value, be it null or null String, it would be much better to load the ImageView with the provided place holder image or just clear the ImageView if no Place Holder is provided by the user. This would be a better Fall back for the Image Loading Right? Can you please include such behaviour.

JakeWharton:
No. See my previous comments as to why.

jeanpul:
Hi,
Is there a particular reason to raised IllegalArgumentException as it's unchecked ? Some webservices returns empty strings so imho it would be preferable to warn the user of the API
Thanks

muzafferyilmaz:
Picasso's behaviour is not wrong but we are not living perfect world. I don't want to check every time and I can't change server side codes. Global setting can be very useful for this
`Picasso.showErrorIfNullOrEmpty(true);`

jeanpul:
I'm agree with you, I did not knew this global settings, this is what I was looking for thanks

lucavitucci:
[...]

PR #1912, opened 2018-05-31
Image path null & empty check + removing empty image path string check test

RahulSDeshpande:
* This update contains the empty string & ```null``` check for image path provided as a parameter to the ```load()``` method.
* ```TextUtils.isEmpty(stringVar)``` will check the whether the provided string object is ```null``` or is of length '0' (zero).
* This will help us avoid throwing of ```IllegalArgumentException``` in case of image path is empty.
* Removing empty image path string check test, as ```Picasso.java``` will be having ```TestUtils.isEmpty(imagePath)``` to validate the empty image path check.


JakeWharton:
This has been requested before (e.g., #899). An empty string is invalid data whereas null is a clear indicator of absence. If you have empty strings coming through your models you should normalize them prior to passing to Picasso so that they're valid URLs or null.

PR #1914, opened 2018-06-01
Remove PrintWriter from API.

NightlyNexus:
dump(buffer(appendingSink(file))) instead of dump(new PrintWriter(buffer(appendingSink(file)).outputStream())

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/StatsSnapshot.java
+++ b/picasso/src/main/java/com/squareup/picasso3/StatsSnapshot.java
@@ -18,8 +18,9 @@ package com.squareup.picasso3;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.Log;
-import java.io.PrintWriter;
-import java.io.StringWriter;
+import java.io.IOException;
+import okio.Buffer;
+import okio.BufferedSink;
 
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso3/StatsSnapshot.java
```diff
@@ -63,47 +64,69 @@
 
   /** Prints out this {@link StatsSnapshot} into log. */
   @SuppressWarnings("UnusedDeclaration") public void dump() {
-    StringWriter logWriter = new StringWriter();
-    dump(new PrintWriter(logWriter));
-    Log.i(TAG, logWriter.toString());
+    Buffer buffer = new Buffer();
+    try {
+      dump(buffer);
+    } catch (IOException e) {
```
jrodbx: propagate the IOException to the caller?

NightlyNexus: There's no I/O happening here, writing to the buffer. It's an impossible case and would force the caller to assert that it can't happen.

jrodbx: got it.  didn't realize Buffer's implementation of BufferSink.writeUtf8 doesn't declare the IOException in its signature.  nvm.

PR #1965, opened 2018-08-16
Disallow creating instances of PicassoProvider.

NightlyNexus:


JakeWharton:
There is an error prone check for this. Let's turn it on.

NightlyNexus:
I can't a check for this. The closest I could find is http://errorprone.info/bugpattern/PrivateConstructorForNoninstantiableModule

I haven't followed EP in a while, so I could have missed it.

NightlyNexus:
Ah, dunno how I missed that. I tried this locally, and it got upset about BuildConfig.java. We can do this when we stop generating that thing.

JakeWharton:
You can also exclude generated stuff:
https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/buildSrc/src/main/kotlin/androidx/build/ErrorProneConfiguration.kt#72

PR #1975, opened 2018-08-19
Add lifecycle awareness to Picasso

### LINKED ISSUE #1972: Add lifecycle pausing / cancelation support

jrodbx:
Closes #1972

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Picasso.java
@@ -15,6 +15,9 @@
  */
 package com.squareup.picasso3;
 
+import android.arch.lifecycle.Lifecycle;
+import android.arch.lifecycle.LifecycleObserver;
+import android.arch.lifecycle.OnLifecycleEvent;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Color;
@@ -59,7 +62,7 @@ import static com.squareup.picasso3.Utils.log;
  * Use {@see PicassoProvider#get()} for a global singleton instance
  * or construct your own instance with {@link Builder}.
  */
-public class Picasso {
+public class Picasso implements LifecycleObserver {
 
   /** Callbacks for Picasso events. */
   public interface Listener {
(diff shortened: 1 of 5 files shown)


picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -231,6 +254,26 @@ public void cancelTag(@NonNull Object tag) {
     }
   }
 
+  @OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)
```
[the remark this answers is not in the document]
jrodbx: good catch, was lazy mapping based on the method names!

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -231,6 +254,26 @@ public void cancelTag(@NonNull Object tag) {
     }
   }
 
+  @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
```
NightlyNexus: have never used this library. super curious. can't figure out why `LifecycleObserver` is a marker interface that tells `Lifecycling` to look up annotated methods reflectively instead of having the interface implement the methods. seems like the worst of both options.

NightlyNexus:
I mostly like that this will just work in Picasso.

What happens if I am using Picasso to download things that aren't tied to my UI? What if I want my Target to get called into even if the app is in the background? Seems like I want a way to opt into this lifecycle stuff.

NightlyNexus:
I guess then I just don't install it into my Activity. Almost wonder if this should be in a separate artifact that exposes a `LifecycleObserver`, but then again everybody uses AppCompat now which brings in everything and then some.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: Tbh, didn't dig into this too much, went by the recommendation here: https://developer.android.com/reference/android/arch/lifecycle/Lifecycle

Looking at the source now though: DefaultLifecycleObserver has default methods (so Java 8 only), but a dummy impl of FullLifecycleObserver would also work on Java 7.

jrodbx:
I'm open to that.  Want to submit a PR?

PR #1982, opened 2018-08-25
Fix recycled Bitmap check.

NightlyNexus:
fixes the sample app crashing at start.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
@@ -429,7 +429,7 @@ class BitmapHunter implements Runnable {
 
       // If the transformation returned a new bitmap ensure they recycled the original.
       if (newResult != result
-          && newResult.getBitmap() != result.getBitmap()
+          && newResult.getBitmap() != bitmap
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -428,7 +428,9 @@ static Result applyTransformations(Picasso picasso, Request data,
       }
 
       // If the transformation returned a new bitmap ensure they recycled the original.
-      if (newResult != result && !bitmap.isRecycled()) {
+      if (newResult != result
+          && newResult.getBitmap() != result.getBitmap()
```
JakeWharton: Left side can just be bitmap, right?

NightlyNexus: right side.

NightlyNexus:
yeah, the matrix transformation. must have been broken before, but no transformations are in the sample app.

jrodbx:
Interesting, the crash seems to only happen on P.  Tested on a Nexus 5X API 25, Pixel 2 API 27, and a Nexus 5X API 28.

PR #1983, opened 2018-08-25
Yank out one of the target wrappers.

NightlyNexus:


picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -428,7 +428,9 @@ static Result applyTransformations(Picasso picasso, Request data,
       }
 
       // If the transformation returned a new bitmap ensure they recycled the original.
-      if (newResult != result && !bitmap.isRecycled()) {
+      if (newResult != result
+          && newResult.getBitmap() != bitmap
+          && !bitmap.isRecycled()) {
```
JakeWharton: Didn't this already land? Rebase?

picasso/src/main/java/com/squareup/picasso3/FetchAction.java
```diff
@@ -15,14 +15,13 @@
...
+import android.support.annotation.NonNull;
 
-  private final Object fetchTarget;
+class FetchAction extends Action {
   private Callback callback;
 
   FetchAction(Picasso picasso, Request data, Callback callback) {
-    super(picasso, null, data);
-    this.fetchTarget = new Object();
+    super(picasso, data);
```
jrodbx: any idea why this used to return `new Object()` vs this?  was it to avoid strong references?

picasso/src/test/java/com/squareup/picasso3/TestUtils.java
```diff
@@ -53,6 +53,7 @@
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
```
jrodbx: avoidable?

picasso/src/test/java/com/squareup/picasso3/DispatcherTest.java
```diff
@@ -263,7 +265,7 @@
   }
 
   @Test public void performErrorCleansUpAndPostsToMain() {
-    BitmapHunter hunter = mockHunter(URI_KEY_1, new RequestHandler.Result(bitmap1, MEMORY));
+    BitmapHunter hunter = mockHunter(URI_KEY_1 + KEY_SEPARATOR, new RequestHandler.Result(bitmap1, MEMORY));
```
jrodbx: why append KEY_SEPARATOR now?

NightlyNexus: Because now the getKey method is not mocked. It uses the real Request's key.
When the key is built with a user-set stable key prefix, it appends the spearator onto the key.

NightlyNexus: All very unfortunate.

JakeWharton: That's the phrase I use to describe almost all of Picasso's tests!

NightlyNexus:
Ping, if somebody finds some time to take a look.

PR #1987, opened 2018-08-26
Add Picasso.newBuilder()

### LINKED ISSUE #1750: Add Picasso.newBuilder()
We have a screen in our app that wants to keep a memory cache of images it loads for as long as we're in the screen's scope and then evict all the screen's Bitmaps from the cache when the scope is torn down.
Right now, there's a very old fork of Picasso that added tags to Requests and then evicts all Bitmaps with the tag from a single cache.

While not exactly the same, a better solution might be to do Picasso.newBuilder() on the existing Picasso instance, replace the cache with a new one, and build() to make a Picasso with a new Dispatcher with a new cache.

Without newBuilder(), I'll need to pass down the Downloader (and any other customizations) manually to this screen's scope creation instead of just the Picasso instance.

Open to a `newBuilder()` API?

jrodbx:
Closes #1750

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Picasso.java
@@ -124,12 +124,13 @@ public class Picasso implements LifecycleObserver {
     }
   };
 
-  @Nullable private final Listener listener;
-  private final List<RequestTransformer> requestTransformers;
-  private final List<RequestHandler> requestHandlers;
+  @Nullable final Listener listener;
+  final List<RequestTransformer> requestTransformers;
+  final List<RequestHandler> requestHandlers;
 
   final Context context;
   final Dispatcher dispatcher;
+  final Call.Factory callFactory;
   private final @Nullable okhttp3.Cache closeableCache;
   final PlatformLruCache cache;
   final Stats stats;
@@ -149,6 +150,7 @@ public class Picasso implements LifecycleObserver {
       boolean loggingEnabled) {
     this.context = context;
     this.dispatcher = dispatcher;
+    this.callFactory = callFactory;
     this.closeableCache = closeableCache;
     this.cache = cache;
     this.listener = listener;
@@ -646,6 +648,11 @@ public class Picasso implements LifecycleObserver {
     }
   }
 
+  @NonNull
+  public Builder newBuilder() {
+    return new Builder(this);
+  }
+
   /** Fluent API for creating {@link Picasso} instances. */
   @SuppressWarnings("UnusedDeclaration") // Public API.
   public static class Builder {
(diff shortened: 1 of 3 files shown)


picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -660,11 +667,30 @@ void cancelExistingRequest(Object target) {
...
+    }
+
+    Builder(Picasso picasso) {
+      context = picasso.context;
+      callFactory = picasso.callFactory;
+      service = picasso.dispatcher.service;
+      cache = picasso.cache;
+      listener = picasso.listener;
+      requestTransformers.addAll(picasso.requestTransformers);
+      requestHandlers.addAll(picasso.requestHandlers);
```
NightlyNexus: won't this end up adding all the default request handlers again?

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -710,7 +736,7 @@ public Builder callFactory(@NonNull Call.Factory factory) {
     @NonNull
     public Builder executor(@NonNull ExecutorService executorService) {
       checkNotNull(executorService, "executorService == null");
-      if (this.service != null) {
+      if (!isChild && this.service != null) {
         throw new IllegalStateException("Executor service already set.");
```
NightlyNexus: remove these checks and the isChild flag. the checks don't add any value.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: yep.  good catch.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: yea, I went back and forth on this.  I guess it made sense before, when passing Builders around?

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: updated.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -660,13 +669,28 @@ void cancelExistingRequest(Object target) {
 
     private boolean indicatorsEnabled;
     private boolean loggingEnabled;
+    private boolean isChild;
```
NightlyNexus: Nix this.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -660,13 +669,28 @@ void cancelExistingRequest(Object target) {
...
     }
 
+    Builder(Picasso picasso) {
+      context = picasso.context;
+      callFactory = picasso.callFactory;
+      service = picasso.dispatcher.service;
+      cache = picasso.cache;
+      listener = picasso.listener;
+      requestTransformers.addAll(picasso.requestTransformers);
+      requestHandlers.addAll(picasso.extraRequestHandlers);
```
NightlyNexus: Nit: could we do a subList call here on the requestHandlers instead of making the extra field? I think that's what Moshi and Retrofit do.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: Nixed.

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: Ok.  This feels like one of those things that will be easy to forget to update..

JakeWharton: Test case then? We have ones in Retrofit at least to ensure the default ones don't get copied twice.

jrodbx: Yup, there's a test case here: https://github.com/square/picasso/blob/aa6005e3115c48b1e411cb1c5a14272b111601b2/picasso/src/test/java/com/squareup/picasso3/PicassoTest.java#L565.

Thoughts?

jrodbx: ping @JakeWharton

PR #1991, opened 2018-09-04
Ignore case for http(s) scheme

### LINKED ISSUE #1990: NetworkRequestHandler cannot handle uppercased HTTP or HTTPS schemes
I'm getting a `BitmapHunter Exception - Unrecognized type of request` even though everything seems fine, as stated in the title, somehow my URL ended up using **HTTP** (UPPERCASE) and here:

https://github.com/square/picasso/blob/6f3189fb94f0c1a647a5ada605c505e00caf51f5/picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.java#L46-L52

Is my (and possible others) problem, i don't know if it's intended to NOT support the UPPERCASE case scheme, but using `(SCHEME_HTTP.equalsIgnoreCase(scheme) || SCHEME_HTTPS.equalsIgnoreCase(scheme))` would solve the problem.

Also, this is a silly mistake that is not easy to find, if the insensitive case is intended, at least put some good exception message..

Grohden:
Could not use [normalizeScheme](https://developer.android.com/reference/android/net/Uri.html#normalizeScheme()) due to api level

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
+++ b/picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
@@ -214,12 +214,8 @@ public class NetworkRequestHandlerTest {
     };
 
     for (String scheme : schemes) {
-      final Uri uri = URI_1.buildUpon().scheme(scheme).build();
-      final Boolean shouldHandle = networkHandler.canHandleRequest(
-              TestUtils.mockRequest(uri)
-      );
-
-      assertThat(shouldHandle).isTrue();
+      Uri uri = URI_1.buildUpon().scheme(scheme).build();
+      assertThat(networkHandler.canHandleRequest(TestUtils.mockRequest(uri))).isTrue();
     }
   }
 
--- a/picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
+++ b/picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
@@ -16,6 +16,7 @@
 package com.squareup.picasso3;
 
 import android.net.NetworkInfo;
+import android.net.Uri;
 import android.support.annotation.NonNull;
 import com.squareup.picasso3.RequestHandler.Result;
 import com.squareup.picasso3.TestUtils.PremadeCall;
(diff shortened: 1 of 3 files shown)


Grohden:
@JakeWharton you are right! also, i'm wondering if only **allowing** network request handler to handle this would solve. I will write tests today

Grohden:
@JakeWharton I wrote the tests (i don't know if they are in following the conventions.. also, this is my first test) and i also tested the sample app with the schema in uppercase, everything loaded correctly.

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
```diff
@@ -203,6 +204,26 @@
...
+            "https",
+            "HTTP",
+            "HTTPS",
+            "HTtP",
+    };
+
+    for (String scheme : schemes) {
+      final Uri uri = URI_1.buildUpon().scheme(scheme).build();
+      final Boolean shouldHandle = networkHandler.canHandleRequest(
+              TestUtils.mockRequest(uri)
```
NightlyNexus: i'd inline the `mockRequest` for tests, but i'm hardly opinionated here.

Grohden: I didn't inline because i would need to import the whole package there, from picasso instead of okhttp.. :/

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
```diff
@@ -203,6 +204,26 @@
...
+    String[] schemes = {
+            "http",
+            "https",
+            "HTTP",
+            "HTTPS",
+            "HTtP",
+    };
+
+    for (String scheme : schemes) {
+      final Uri uri = URI_1.buildUpon().scheme(scheme).build();
```
jrodbx: nit: remove final

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
```diff
@@ -203,6 +204,26 @@
...
+      final Boolean shouldHandle = networkHandler.canHandleRequest(
```
jrodbx: no final, boolean

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
```diff
@@ -203,6 +204,26 @@
...
+              TestUtils.mockRequest(uri)
+      );
+
+      assertThat(shouldHandle).isTrue();
```
jrodbx: can probably inline here.

jrodbx:
Fixes #1990

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.java
Grohden: i didn't inline here for legibility and column limit (that's why i used final btw, i always think that the compiler will inline "one reference/usage final" variables - but i know that even without `final` the compiler can do that), anyway, i will inline it

PR #2001, opened 2018-09-11
Wrap Source to prevent propagating bad Bitmaps.

### LINKED ISSUE #1998: Use exception-catching Source wrapper to prevent propagating bad Bitmaps
Sometimes Bitmap.decodeStream doesn't throw when the underlying stream does. This is unequivocally an Android bug. However, we can trivially detect this case by wrapping the source, marking any exceptions that are thrown, and re-throwing if the method erroneously returns normally.

NightlyNexus:
Closes #1998

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
+++ b/picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
@@ -30,10 +30,13 @@ import android.util.TypedValue;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
+import okio.Buffer;
 import okio.BufferedSource;
+import okio.ForwardingSource;
 import okio.Okio;
 import okio.Source;
 
+import static android.os.Build.VERSION.SDK_INT;
 import static com.squareup.picasso3.Utils.checkNotNull;
 
 final class BitmapUtils {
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
```diff
@@ -196,6 +199,29 @@ static boolean isXmlResource(Resources resources, @DrawableRes int drawableId) {
...
+    IOException thrownException;
+
+    ExceptionCatchingSource(Source delegate) {
+      super(delegate);
+    }
+
+    @Override public long read(Buffer sink, long byteCount) throws IOException {
+      try {
+        return super.read(sink, byteCount);
+      } catch (IOException e) {
```
NightlyNexus: i'm assuming BitmapFactory.decodeStream doesn't swallow other throwables...

NightlyNexus:
It'd be nice to log or something when this happens. I'd like to know how often and on what devices this happens.

JakeWharton:
Sure. Add it!

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
```diff
@@ -93,14 +95,12 @@ static void calculateInSampleSize(int reqWidth, int reqHeight, int width, int he
    * about the supplied request in order to do the decoding efficiently (such as through leveraging
    * {@code inSampleSize}).
    */
-  static Bitmap decodeStream(Source source, Request request) throws IOException {
-    BufferedSource bufferedSource = Okio.buffer(source);
-
+  static Bitmap decodeStream(BufferedSource source, Request request) throws IOException {
```
JakeWharton: Why did you make this change? Since we're using a Source to wrap the incoming source it makes this buffering layer mostly useless. Why not wrap the underlying Source as close as possible to the original before applying any buffering layer?

NightlyNexus:
yep, agree on both comments. will return!

JakeWharton:
Thanks!

PR #2011, opened 2018-10-01
 Replace SourceBufferingInputStream with BufferedSource.peek().

NightlyNexus:


build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
     'sourceCompatibility': JavaVersion.VERSION_1_7,
     'targetCompatibility': JavaVersion.VERSION_1_7,
     'okhttp': '3.11.0',
-    'okio': '1.15.0',
+    'okio': '2.1.0',
```
JakeWharton: If we do this, we can just convert all of Picasso to Kotlin as well. So... let's do that.

PR #2014, opened 2018-10-06
A base LRU cache that allows you to store EXIF data with the Bitmap.

extorn:
A base LRU cache that allows you to store EXIF data (or whatever else you want) alongside the bitmap.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/BaseLruExifCache
+++ b/picasso/src/main/java/com/squareup/picasso3/BaseLruExifCache
@@ -1,0 +1,238 @@
+package com.squareup.picasso;
+
+import android.content.Context;
+import android.graphics.Bitmap;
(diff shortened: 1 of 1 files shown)


extorn:
You may use the code as you wish. There are no tests else I'd have supplied them.
I didn't really think you'd have interest in this else it would have been done already. That said I maintain a cache of image metadata linked to an image in this way to for example to avoid having to rescan the image. The purpose of the PR was to raise a suggestion with example rather than as something that could be committed.

NightlyNexus:
Hey, thanks for the code. Also, feel free to open issues with feature requests, too.

We don't have the Cache interface anymore. I think you can achieve this independently of Picasso by keeping your own metadata cache around. Is there anything Picasso needs to do to support your use case?

Issue #2018, opened 2018-10-13
Use Okio 1.16.0 not 2.1.0


JakeWharton:
Which means we're not forced into Kotlin 😢

jrodbx:
Pikasso!

PR #2034, opened 2018-11-08
Change okio version to 1.16.0

Ramkishorevit:
Changed the okio version as per #2018


jrodbx:
Thanks for the contribution, but we'll probably migrate to Kotlin for a 3.0 release or a fast follow.

Issue #2007, opened 2018-09-25
Abstract Request


NightlyNexus:
Requests can have different properties, and different RequestHandlers can use or not use those properties.

For example, network Requests have HTTP headers. (Picasso 2 has no API to add a request header to a single request.)

Request needs to support variable property types, possibly via type inheritance.

NightlyNexus:
yeah, a tag or map would work.

JakeWharton:
I like a Map because it will have symmetry with #639

Ramkishorevit:
@NightlyNexus It would be great to have HTTP headers support for the network requests. Could you please highlight some more examples of different properties apart from the HTTP headers? So that I can get more insight on implementing this. Also I am assuming the equation here is to represent the headers as a single Map.

NightlyNexus:
`@swankjesse` suggested having an API just for HTTP headers. Then, any other non-HTTP usages/handers/interceptors could read data off of custom headers (rare case), similar to how data can be passed into OkHttp Interceptors currently. I am in favor of that.

Ramkishorevit:
@NightlyNexus so you mean something like this,

`Picasso.with(context).load(uri).addHeader(map).into(view);`

But personally I feel header is a property that has more closer association with the Uri. Wouldn't it be good if we do something like,

`PicassoUri uri = new PicassoUri(url, headerMap);
  Picasso.with(context).load(uri).into(view);`
Just a thought. I like both anyways. Please let me know your thoughts on this.

JakeWharton:
The headers do not affect things like caching which your snippet implies by
combining it with the Uri. Plus I don't want to invent another type for
carrying composite data–the request is already that type.

Ramkishorevit:
@JakeWharton Thanks. That makes a lot of sense.

Issue #639, opened 2014-08-30
Add metadata into the pipeline


ghost:
Android L has a new feature that will extract common colors out of an image. Palette.generate(bitmap) This should be run on a background thread. I think it would be cool and efficient to provide a callback after the image has been loaded into memory(from disk or net) on the SAME thread so we can do more processing.  I don't think transitions are the correct place for this but I could be wrong and they also don't rerun transition on loading from disk which i would need it to do. I need to call that function and add the results to an internal cache for fast retrieval for activity animations.

I know i could use the Target class and kick off a new task to compute this but we already have thread that had to deal with the bitmap. Just seems inefficient to spin these up and I feel like this is going to need to be done a lot in the near future.

Also this is more of a request. I have no idea if this is the correct place for this.

Thanks
Joe

dnkoutso:
Yes this will be done using generic means of metadata for requests.

PR #2035, opened 2018-11-10
API to add request headers #2007

Ramkishorevit:
Implementation of request headers as per the discussion - #2007 

Usage is something like,
```
PicassoProvider.get().load(uri).addHeaders(map).into(view);
```

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Request.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Request.java
@@ -19,19 +19,17 @@ import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Looper;
 import android.view.Gravity;
-
-import com.squareup.picasso3.Picasso.Priority;
-
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-
 import androidx.annotation.DrawableRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.Px;
 import okhttp3.Headers;
+
+import com.squareup.picasso3.Picasso.Priority;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import static com.squareup.picasso3.Utils.MAIN_THREAD_KEY_BUILDER;
 import static com.squareup.picasso3.Utils.checkNotNull;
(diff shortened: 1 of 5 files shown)


Ramkishorevit:
@NightlyNexus @JakeWharton Could you guys please review this?

JakeWharton:
On our own schedule, yes.

NightlyNexus:
Let's use okhttp3.Headers instead of a map. That way, we can represent HTTP headers correctly, have clear immutability, and have add instead of only set operations.
Maybe the Picasso request builder type should expose the Headers.Builder directly for simplicity.

Ramkishorevit:
@NightlyNexus Immutability wise okhttp3.Headers appears to be better. But I thought map would be better as it would have symmetry with #639 as @JakeWharton mentioned. Also I feel map will be well known type for an end-user compared to exposing okhttp3.Headers. What do you think?

JakeWharton:
One difference is the metadata is untyped key-value pairs whereas headers
are always String to String in (basically) ASCII as a multimap. I'm not
sure we should support untyped values for requests.

Ramkishorevit:
@JakeWharton Maybe we shouldn't actually allow untyped values for requests. Then I guess okhttp3.Headers will be a better option. It appears to be more simpler providing immutability & also looks typed. Shall I go ahead and change it to okhttp3.Headers rather than map?

JakeWharton:
Please!

Ramkishorevit:
Changed to okhttp3.Headers.

picasso/src/main/java/com/squareup/picasso3/Request.java
```diff
@@ -640,6 +647,12 @@ public Builder networkPolicy(@NonNull NetworkPolicy policy,
       return this;
     }
 
+    @NonNull
+    public Builder addHeaders(@Nullable Headers headers) {
```
NightlyNexus: i think we'll eventually want to make it easy to add and remove headers.
like exposing `public Headers.Builder Request.Builder.headers`.

and then the Headers at `Request.headers` is never null.

Ramkishorevit: @NightlyNexus should i change this method so that it exposes Headers.Builder ? 
Like,  public Headers.Builder addHeaders(..) ?

Ramkishorevit: @NightlyNexus @JakeWharton Could you please give me clarification on this?

JakeWharton: No. The method should take two `String`s. Most people don't care how the headers are actually represented.

Ramkishorevit: @JakeWharton @NightlyNexus updated the method to collect two Strings.

PR #2040, opened 2018-11-24
Fixing broken download link in website

GabrielBB:
Creating this PR because [this one](https://github.com/square/picasso/pull/1974) only fixed README.md but not the website. For this fix i'm updating JQuery Maven Artifact Plugin to version 2.0.0 to make use of the "packaging" property so it looks for an AAR instead of a JAR.


JakeWharton:
Thanks!

website/index.html
```diff
@@ -196,12 +196,16 @@ <h3 id="license">License</h3>
         });
 
         // Look up the latest version of the library.
-        $.fn.artifactVersion('com.squareup.picasso3', 'picasso', function(version, url) {
-          $('.version').text(version);
-          $('.version-tag').text('v' + version);
-          $('.version-href').attr('href', url);
+        $.fn.artifactVersion({
+          'groupId': 'com.squareup.picasso',
```
jrodbx: shouldn't this be `picasso3`?

GabrielBB: When i cloned the project it had picasso3 and it wasn't working. I didn't know why until i checked the live code in the deployed website and it had "picasso". When i used picasso then the "artifactVersion" function worked. I don't know why the repo had a different code, tho

PR #2041, opened 2018-11-26
Fix groupId typo in website html

jrodbx:
Follow-up to https://github.com/square/picasso/pull/2040


NightlyNexus:
don't we want it to point to Picasso e (which is com.squareup.picasso)?

JakeWharton:
Changing this prevents deploying the website until release. This shouldn't be changed until the readme is also changed

PR #2043, opened 2018-11-28
Center debug indicators

### LINKED ISSUE #791: Possibly center debug indicators?
In too many places picasso is used with centerCrop() which makes debug indicators invisible (cropped out). And one more scenario: using RoundedImageView also hides the indicator. Having it in the center would fix these kinds of problems. Just making my case :)

ouattararomuald:
Center debug indicators and closes #791.


k-ozam:
Need this feature. It seems some rounded ImageView can't display the indicator.

PR #2056, opened 2019-02-04
Add lint rules to pollexor artifact.

NightlyNexus:


JakeWharton:
Should we have global rules for all modules?

NightlyNexus:
#2059

PR #2060, opened 2019-02-05
Let high priority images render sooner.

### LINKED ISSUE #2026: Option to post HUNTER_COMPLETE at the front of the message queue?
At startup, our app gets dozens of messages backed up on our message queue in the main thread handler.
This can cause 100+ ms of delay in delivering the HUNTER_COMPLETE message to show our first images.
If Picasso continues using the message queue, maybe it should have a way to mark the images as the top priority?

NightlyNexus:
High priority images get complete messages sent at the front of the Handler message queue.
Closes #2026

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Dispatcher.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Dispatcher.java
@@ -421,7 +421,12 @@ class Dispatcher {
       }
     }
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
Should we document this behavior somewhere? On the priority method perhaps

NightlyNexus:
good idea. #2061

PR #2062, opened 2019-02-07
Fix non-Bitmap results not continuing transforms.

### LINKED ISSUE #2055: Remove check for recycled original bitmap from BitmapHunter
We started seeing this crash but only on Huawei Mate 20 Pro running Android 9. Could this be an issue with those specific devices or simply means we run out of memory?

Fatal Exception: java.lang.IllegalStateException: Transformation max-width-190 mutated input Bitmap but failed to recycle the original.
       at com.squareup.picasso.BitmapHunter$6.run(BitmapHunter.java:490)

Picasso version: 2.71828
Picasso Transformations version: 2.1.2

NightlyNexus:
Also fixes allowing recycled Bitmaps to be returned in transformations and removes unnecessary check for recycling old Bitmaps.

Closes #2055.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
+++ b/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
@@ -418,29 +418,13 @@ class BitmapHunter implements Runnable {
         return null;
       }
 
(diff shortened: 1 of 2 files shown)


jrodbx:
👍

Issue #2065, opened 2019-02-15
Delete Picasso.Builder.threadFactory.


NightlyNexus:
I added this API to set the thread priorities to foreground easily for the default, built-in PicassoExecutorService, but I need to supply my own full executor, anyway. I suspect others who care about such things will need to tweak a full implementation, too.

PR #2068, opened 2019-02-26
Expose PicassoExecutorService.

### LINKED ISSUE #2064: Custom executor cannot maintain request priority ordering.
The default ThreadPoolExecutor maxes at 3 threads. From production experimentation, the app I work on has found 4 threads to be best for our use case (typically showing 4 images on our home screen).
However, there is no way from the API to supply my own custom executor and maintain request priority ordering because that relies on Picasso's [implementation details](https://github.com/square/picasso/blob/89f55b76e3be2b65e5997b7698f782f16f8547e3/picasso/src/main/java/com/squareup/picasso3/PicassoExecutorService.java#L41).
I actually don't use this request priority ordering (I do HIGH on all for the postToFrontOfQueue optimization), but this could be a general problem.

NightlyNexus:
Supplying a custom executor currently means not being able to use Request priority, since there is currently no way to get that information from the internal BitmapHunter in a custom executor.
Exposing PicassoExecutorService lets users tweak settings like the thread count and thread priority easily.
Closes #2064 and #2065.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/PicassoExecutorService.java
+++ b/picasso/src/main/java/com/squareup/picasso3/PicassoExecutorService.java
@@ -15,6 +15,8 @@
  */
 package com.squareup.picasso3;
 
+import android.os.Process;
+import androidx.annotation.NonNull;
 import java.util.concurrent.Future;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.PriorityBlockingQueue;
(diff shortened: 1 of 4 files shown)


picasso/src/main/java/com/squareup/picasso3/PicassoExecutorService.java
```diff
@@ -15,34 +15,67 @@
...
+import static android.os.Process.THREAD_PRIORITY_BACKGROUND;
+
 /**
  * The default {@link java.util.concurrent.ExecutorService} used for new {@link Picasso} instances.
- * <p>
- * Exists as a custom type so that we can differentiate the use of defaults versus a user-supplied
- * instance.
  */
-class PicassoExecutorService extends ThreadPoolExecutor {
+public class PicassoExecutorService extends ThreadPoolExecutor {
```
NightlyNexus: can't make this final without rewriting a bunch of Dispatcher tests.

PR #2076, opened 2019-03-21
Fix ImageDecoder stretching instead of cropping.

NightlyNexus:
Make ImageDecoder retain the image aspect ratio, so that cropping can be applied after decoding.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
+++ b/picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
@@ -75,26 +75,30 @@ final class BitmapUtils {
         || (targetHeight != 0 && inHeight > targetHeight);
   }
 
-  static void calculateInSampleSize(int reqWidth, int reqHeight, int width, int height,
+  static void calculateInSampleSize(int requestWidth, int requestHeight, int width, int height,
       BitmapFactory.Options options, Request request) {
-    int sampleSize = 1;
-    if (height > reqHeight || width > reqWidth) {
-      final int heightRatio;
-      final int widthRatio;
-      if (reqHeight == 0) {
-        sampleSize = (int) Math.floor((float) width / (float) reqWidth);
-      } else if (reqWidth == 0) {
-        sampleSize = (int) Math.floor((float) height / (float) reqHeight);
-      } else {
-        heightRatio = (int) Math.floor((float) height / (float) reqHeight);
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
```diff
@@ -75,28 +75,32 @@ static boolean shouldResize(boolean onlyScaleDown, int inWidth, int inHeight,
...
+    options.inSampleSize = ratio(requestWidth, requestHeight, width, height, request);
     options.inJustDecodeBounds = false;
   }
 
+  static int ratio(int requestWidth, int requestHeight, int width, int height, Request request) {
+    if (height <= requestHeight && width <= requestWidth) {
+      return 1;
+    }
+    if (requestHeight == 0) {
+      int widthRatio = width / requestWidth;
```
JakeWharton: You can move this above the `if` since we need to compute it no matter what.

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
```diff
@@ -75,28 +75,32 @@ static boolean shouldResize(boolean onlyScaleDown, int inWidth, int inHeight,
...
+      return widthRatio != 0 ? widthRatio : 1;
+    }
+    if (requestWidth == 0) {
+      int heightRatio = height / requestHeight;
```
JakeWharton: Similarly, move before this `if`

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
```diff
@@ -75,28 +75,32 @@ static boolean shouldResize(boolean onlyScaleDown, int inWidth, int inHeight,
...
+      return heightRatio != 0 ? heightRatio : 1;
+    }
+    int heightRatio = height / requestHeight;
+    int widthRatio = width / requestWidth;
```
JakeWharton: Then nuke these 2 lines

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.java
NightlyNexus: if requestHeight != 0 and requestWidth == 0, then we don't need to compute widthRatio, though.

JakeWharton: True. Let's leave it as is.

PR #2103, opened 2019-08-17
okhttp 4.1.0

sullis:


jrodbx:
That's fine.  3.0 will include converting to Kotlin APIs.

PR #2104, opened 2019-08-28
fix: NPE when load resource drawable

tRuNKator:


joshschriever:
@JakeWharton @jrodbx could we get some eyes on this? I'm getting this NPE when trying to load resource drawables.

jrodbx:
I'll take a look this weekend.

PR #2109, opened 2019-09-11
Konvert some sample to Kotlin

jrodbx:
See first commit for kdiff


jrodbx:
Yep, def in need of some love.

PR #2111, opened 2019-09-20
Temporarily delete Picasso tests

jrodbx:
Current fixtures overuse mocking and make feature development harder.


JakeWharton:
A lot of these tests are extremely high-value and the incentive to add them back in a different form will be extremely low if they are deleted. I would rather see them all retained either in an `src/oldTests/` directory which can be slowly burned down through migration over time.

jrodbx:
I feel that, how about simply @Ignore'ing them then?

jrodbx:
nvm, then compilation will fail

PR #2117, opened 2019-10-01
Reduce multiple getLooper calls

jrodbx:
This is less of a micro-optimization than a simple workaround to what appears to be a bug in layoutlib, which results in Picasso crashing on creation in Paparazzi tests.  The issue: subsequent calls to `HandlerThread.getLooper()` fail, due to change in `Thread.isAlive()`'s return value.

See https://github.com/cashapp/paparazzi/compare/1524f4c...jrod/2019-09-30/handler-thread-bug for an example test comparison.

as paparazzi test:
```
state: RUNNABLE
isAlive: true
looper: Looper (testThread, tid 17) {121314f7}
state: TERMINATED
isAlive: false
looper: null
state: TERMINATED
isAlive: false
```

as instrumentation test:
```
state: RUNNABLE
isAlive: true
looper: Looper (testThread, tid 612) {9d4c42b}
state: RUNNABLE
isAlive: true
looper: Looper (testThread, tid 612) {9d4c42b}
state: RUNNABLE
isAlive: true
```

Filed as https://issuetracker.google.com/issues/141925961

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Dispatcher.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Dispatcher.java
@@ -95,14 +95,15 @@ class Dispatcher {
       PlatformLruCache cache, Stats stats) {
     this.dispatcherThread = new DispatcherThread();
     this.dispatcherThread.start();
-    Utils.flushStackLocalLeaks(dispatcherThread.getLooper());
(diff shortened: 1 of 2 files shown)


jrodbx:
Updated original comment with link to bug on Google's issuetracker.

PR #2120, opened 2019-10-24
Use https in one-liner intro example

BriChri:
The http example won't load on Android 9 using the default network security configuration.
https://developer.android.com/training/articles/security-config.html#CleartextTrafficPermitted


jrodbx:
thanks!

PR #2133, opened 2019-12-12
PlatformLruCache to Kotlin

jrodbx:


picasso/src/main/java/com/squareup/picasso3/PlatformLruCache.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.core.graphics.BitmapCompat
+
+typealias Size = Int
+
+/** A memory cache which uses a least-recently used eviction policy.  */
+internal class PlatformLruCache(maxByteCount: Int) {
+
+  /** Create a cache with a given maximum size in bytes.  */
+  val cache =
+    object : LruCache<String, Pair<Bitmap, Size>>(if (maxByteCount != 0) maxByteCount else 1) {
```
JakeWharton: Use of a `Pair` causes boxing. We're already allocating the `Pair`, might as well just go back to allocating our own type which won't box the size.

jrodbx: Makes sense.  Updated.

PR #2135, opened 2019-12-12
Actions to Kotlin

jrodbx:


picasso/src/main/java/com/squareup/picasso3/Action.kt
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.picasso3
+
+import com.squareup.picasso3.RequestHandler.Result
+
+abstract class Action(
```
JakeWharton: internal?

picasso/src/main/java/com/squareup/picasso3/BitmapTargetAction.kt
```diff
@@ -0,0 +1,49 @@
...
+      target.onBitmapLoaded(bitmap, result.loadedFrom)
+      check(!bitmap.isRecycled) { "Target callback must not recycle bitmap!" }
+    }
+  }
+
+  override fun error(e: Exception) {
+    if (errorResId != 0) {
+      target.onBitmapFailed(e, ContextCompat.getDrawable(picasso.context, errorResId))
+    } else {
+      target.onBitmapFailed(e, errorDrawable)
```
JakeWharton: extract common expression from these branches so the `if` only applies to the last argument.

jrodbx: updated

PR #2137, opened 2019-12-16
Convert RequestHandler

jrodbx:
Also, splits Result into DrawableResult and BitmapResult subtypes


picasso/src/main/java/com/squareup/picasso3/RequestHandler.kt
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Returns the resulting [Picasso.LoadedFrom] generated from a [load] call.
+     */
+    @JvmField val loadedFrom: LoadedFrom,
+    /**
+     * Returns the resulting EXIF rotation generated from a [load] call.
+     */
+    @JvmField val exifRotation: Int = 0
+  ) {
+    class BitmapResult @JvmOverloads constructor(
```
JakeWharton: I think we either drop the `Result` suffix if these are nested or we make them sibling classes and keep the suffix. But not nested and with a suffix since it just look redundant.

picasso/src/main/java/com/squareup/picasso3/BitmapHunter.java
```diff
@@ -102,11 +102,13 @@
 
       result = hunt();
 
-      if (result.getBitmap() == null && result.getDrawable() == null) {
-        dispatcher.dispatchFailed(this);
-      } else {
-        dispatcher.dispatchComplete(this);
-      }
+      // TODO can this ever happen anymore?
```
JakeWharton: Nope

picasso/src/main/java/com/squareup/picasso3/RequestHandler.kt
jrodbx: going with nested and dropped Suffix

Issue #2167, opened 2020-05-11
BitmapHunter always delay fail by 500ms even when there should be no retries


fruxo:
How to reproduce: 
-Use Picasso.load(File) to make a request for a file that's missing (or that would throw some other IOException).

Result: 
The result callback fires after 500ms, even though there should be no retries for those requests (its RequestHandler.getRetryCount() returns 0).

Reason:
BitmapHunter.run catches IOException and schedules a retry after RETRY_DELAY. After that delay it checks if there's any retries and if not it calls the result callback.
I would expect it to check first if there's any retries, like this:

      if (retryCount > 0) {
        dispatcher.dispatchRetry(this);
      } else {
        dispatcher.dispatchFailed(this);
      }

That would speed up the request by 500ms.

JakeWharton:
Sounds good. Feel free to send a PR if you want.

PR #2168, opened 2020-05-12
Dispatch fail immediately if there's no retries.

fruxo:
Removes unnecessary 500ms delay for requests that has no retries.

Fixes issue #2167


JakeWharton:
Thanks1

PR #2176, opened 2020-08-03
Migrate 2.x branch to AndroidX

jrodbx:
Plan is to cut another 2.x release, maybe `2.7182818284`?  😄 

Also,

* bumps AGP, compileSdk and targetSdk to latest
* bumps Gradle and Robolectric to more recent releases (opted not for latest to minimize build fixes)


jrodbx:
Wasn't planning on changing the (groupId, package), only the version (2.7182818284?).  My thinking is unblocking consumers wanting to rid Jetifier from their builds, would be an OK tradeoff with the atypical breaking change.

jrodbx:
This is based off the `2.e` branch (not latest which has moved on to `picasso3`) in case that wasn't clear.

JakeWharton:
Yeah I see the branch, I'm just stating that it's technically a binary-incompatible change which would normally require such operations. I guess we can just release it as 2.8 since anyone that's stuck around already dealt with the binary incompatibility of 2.7.

Toxquimauricio (approved):
~~~~

PR #2183, opened 2020-08-31
Update Picasso.java

umerov1999:
Fix exception if URI null


JakeWharton:
When would this occur?

umerov1999:
if url link is null in load request

JakeWharton:
We do not allow null values.

jrodbx:
Thanks for the contribution, but WJWS!

PR #2184, opened 2020-08-31
Fix Synthetic Accessor lint for BitmapHunter.kt file

amilcar-andrade:


amilcar-andrade:
@jrodbx let me know if fixing the linting is something we would want to do, 🙏

jrodbx (approved):

Issue #2189, opened 2020-12-14
Use AppStartup instead of PicassoContentProvider


yogurtearl:
Can use androidx AppStartup instead of `PicassoContentProvider`
https://developer.android.com/topic/libraries/app-startup

This is better for start up time and allows predictable ordering of library initializers.

JakeWharton:
Didn't we remove the content provider? If not, we should. Static singletons are a 2014 convenience and a 2020 burden.

yogurtearl:
The content provider is still there on `master` didn't look to see if/how it is still used.
https://github.com/square/picasso/blob/master/picasso-sample/src/main/java/com/example/picasso/provider/PicassoContentProvider.kt

yogurtearl:
oh, ha, just realized that is in the sample app. 😂  maybe update the sample to use androidx AppStartup?

JakeWharton:
Yep! Would make for a much better copy/paste pattern, if one decides to go that route.

PR #2195, opened 2021-01-31
Use AppStartup to initialize Picasso

bdpqchen:
I noticed that Picasso has used ContentProvider to get the application context, but it is removed from the library now. As this issue #2189 said, it may be a nice thing to use AppStartup now.

This PR uses AppStartup to save the application context object, which is used when creating a static singleton Picasso.

Note:
Although AppStartup reduces the costs when initializing the ContentProvider one by one, but it uses reflection to create the Initializer object (PicassoInitializer), so it worth to discuss the risks and benefits of this PR.


JakeWharton:
We no longer want to provide a static instance out of the box. If you need one, copy this code into your application.

Issue #2158, opened 2020-03-04
Picasso callback onError(Exception ex) throws NullPointerException for Exception object


rajunigadi:
Picasso callback onError throws NullPointerException for exception object though we are not using Exception object anywhere.

Stacktrace:
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.Exception.getMessage()' on a null object reference
        at com.squareup.picasso.Picasso.deliverAction(Picasso.java:583)
        at com.squareup.picasso.Picasso.complete(Picasso.java:528)
        at com.squareup.picasso.Picasso$1.handleMessage(Picasso.java:122)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)

PR #2196, opened 2021-01-31
Fix invoking on a nullable exception object in Picasso#deliverAction

bdpqchen:
I did not reproduce this crash #2158 , but found the code at `Picasso#deliverAction(Exception e)` which `e` is a **Nullable** parameter and use `checkNotNull` to check the parameter `e` , it is the main cause.

This PR create new RuntimeException when parameter `e` is a null value.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Picasso.java
+++ b/picasso/src/main/java/com/squareup/picasso3/Picasso.java
@@ -623,7 +623,7 @@ public class Picasso implements LifecycleObserver {
         log(OWNER_MAIN, VERB_COMPLETED, action.request.logId(), "from " + result.loadedFrom);
       }
     } else {
-      Exception exception = checkNotNull(e, "e == null");
+      Exception exception = (e != null) ? e : new RuntimeException("e == null");
       action.error(exception);
       if (loggingEnabled) {
         log(OWNER_MAIN, VERB_ERRORED, action.request.logId(), exception.getMessage());


JakeWharton:
I would prefer to analyze the paths to this code and determine the underlying cause of the problem rather than ignoring it.

bdpqchen:
The version 2.71828 has this issue and version 2.8 fixed

version 2.71828 : 


```
BitmapHunter#run()
result = hunt();   // ① hunt() will return null bitmap and null exception when loading from a contact record which has no photo.
if (result == null) {
  dispatcher.dispatchFailed(this); // ② dispatch the null bitmap and null exception failed
} else {
  dispatcher.dispatchComplete(this);
}

ContactsPhotoRequestHandler#getInputStream()
case ID_CONTACT:
  return openContactPhotoInputStream(contentResolver, uri, true); // ③ return a null of InputStream

// and then Picasso#deliverAction(Bitmap, _,_, Exception)
action.error(e);
if (loggingEnabled) {
  log(OWNER_MAIN, VERB_ERRORED, action.request.logId(), e.getMessage()); // ④ e == null
}

```

version 2.8 and now
```
ContactsPhotoRequestHandler#getSource() // throw an Exception when could not read a Source
 ?: throw FileNotFoundException("can't open input stream, uri: $uri")
```
It has been fixed in version 2.8, and the readme should update to 2.8, then close that Issue and this PR.

JakeWharton:
There are no plans for any more 2.x releases. There are currently no plans to release master either, however. If you need this fix the best course of action is maintaining a fork.

PR #2267, opened 2022-03-02
Konvert NetworkRequestHandler

jrodbx:


picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.kt
```diff
@@ -0,0 +1,129 @@
...
+          builder.noCache()
+        }
+        if (!shouldWriteToDiskCache(networkPolicy)) {
+          builder.noStore()
+        }
+        builder.build()
+      }
+    }
+
+    val uri = checkNotNull(request.uri, "request.uri == null")
```
Egorand, line 111: nit: use Kotlin's native `checkNotNull` that doesn't allocate the error message String unless it's needed?

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.kt
```diff
@@ -0,0 +1,213 @@
...
+      requests.add(request)
+      try {
+        PremadeCall(request, responses.takeFirst())
+      } catch (e: InterruptedException) {
+        throw AssertionError(e)
+      }
+    }
+  }
+
+  @Test @Throws(Exception::class)
```
Egorand, line 70: nit: are these `@Throws` annotations useful? In Java I believe they were needed to satisfy the compiler, but Kotlin doesn't care. Or are they worth keeping for documentation purposes?

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.kt
```diff
@@ -0,0 +1,213 @@
...
+    hunter.hunt()
+    assertThat(requests.takeFirst().cacheControl.toString())
+      .isEqualTo(CacheControl.FORCE_CACHE.toString())
+  }
+
+  @Test fun shouldRetryTwiceWithAirplaneModeOffAndNoNetworkInfo() {
+    val action = TestUtils.mockAction(URI_KEY_1, URI_1)
+    val cache = PlatformLruCache(0)
+    val hunter = BitmapHunter(picasso, dispatcher, cache, action, networkHandler)
+    assertThat(hunter.shouldRetry(false, null)).isTrue()
```
Egorand, line 106: nit: if possible, using named args here would help improve readability

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+  @Test fun shouldHandleSchemeInsensitiveCase() {
+    val schemes = arrayOf("http", "https", "HTTP", "HTTPS", "HTtP")
+    for (scheme in schemes) {
+      val uri = URI_1.buildUpon().scheme(scheme).build()
+      assertThat(networkHandler.canHandleRequest(TestUtils.mockRequest(uri))).isTrue()
+    }
+  }
+
+  private companion object {
```
Egorand, line 203: nit: doesn't really have to be inside a companion

Egorand (approved):

picasso/src/test/java/com/squareup/picasso3/NetworkRequestHandlerTest.kt
jrodbx: good point! i'll clean that up along with @JvmStatic, etc. once everything is converted.

PR #2269, opened 2022-03-03
Don't mock Picasso in tests

jrodbx:
Replace mock with real Picasso, and where possible, Robolectric context.
Also, increases testing confidence when konverting Picasso itself.

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso3/TestUtils.kt
+++ b/picasso/src/test/java/com/squareup/picasso3/TestUtils.kt
@@ -137,13 +137,15 @@ internal object TestUtils {
   @JvmStatic fun mockRequest(uri: Uri): Request = Request.Builder(uri).build()
 
   @JvmStatic @JvmOverloads fun mockAction(
+    picasso: Picasso,
     key: String,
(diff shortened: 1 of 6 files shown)


picasso/src/test/java/com/squareup/picasso3/TestUtils.kt
```diff
@@ -235,13 +238,15 @@ internal object TestUtils {
...
-  @JvmStatic fun mockPicasso(requestHandler: RequestHandler): Picasso {
-    val picasso = mock(Picasso::class.java)
-    `when`(picasso.getRequestHandlers()).thenReturn(listOf(requestHandler))
-    return picasso
+  @JvmStatic fun mockPicasso(context: Context, requestHandler: RequestHandler): Picasso {
+    return Picasso.Builder(context)
+      .callFactory(UNUSED_CALL_FACTORY)
+      .withCacheSize(0)
+      .addRequestHandler(requestHandler)
+      .build()
```
jrodbx, line 249: where it all starts

chris-horner (approved):
Nice one

PR #2270, opened 2022-03-03
Remove remaining mocks of Picasso

jrodbx:


picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -628,7 +628,7 @@ private void deliverAction(@Nullable Result result, Action action,
     }
   }
 
-  void cancelExistingRequest(Object target) {
+  private void cancelExistingRequest(Object target) {
```
jrodbx, line 631: never needed to be package-private

picasso/src/test/java/com/squareup/picasso3/DeferredRequestCreatorTest.kt
```diff
@@ -109,25 +101,23 @@ class DeferredRequestCreatorTest {
...
     request.cancel()
     verify(target.viewTreeObserver, never()).removeOnPreDrawListener(request)
   }
 
   @Test fun preDrawSubmitsRequestAndCleansUp() {
-    val picasso = mock(Picasso::class.java)
-    `when`(picasso.transformRequest(any(Request::class.java))).thenAnswer(TRANSFORM_REQUEST_ANSWER)
-
-    val creator = RequestCreator(picasso, TestUtils.URI_1, 0)
+    val spyPicasso = spy(picasso) // ugh
```
chris-horner, line 119: 😅

chris-horner (approved):

PR #2272, opened 2022-03-03
Konvert Picasso

jrodbx:
100% Kotlin!


build.gradle
```diff
@@ -29,7 +29,7 @@ buildscript {
     errorProneJavac: 'com.google.errorprone:javac:9+181-r4173-1',
     junit: 'junit:junit:4.13.2',
     truth: 'com.google.truth:truth:1.1.3',
-    robolectric: 'org.robolectric:robolectric:4.4',
+    robolectric: 'org.robolectric:robolectric:4.6.1',
```
jrodbx, line 32: Bitmap transformations weren't copying BitmapConfig over.  Bumping Robolectric fixed that.

picasso/src/test/java/com/squareup/picasso3/TestUtils.kt
```diff
@@ -408,4 +409,21 @@ internal object TestUtils {
...
+  }
+
+  class KArgumentCaptor<T>(
+    private val captor: ArgumentCaptor<T>,
+  ) {
+    val value: T
+      get() = captor.value
+
+    fun capture(): T = captor.capture()
+  }
```
jrodbx, line 428: I didn't want to add yet another Mockito dependency, so this unblocks.

gamepro65 (approved):

picasso/src/main/java/com/squareup/picasso3/Picasso.java
```diff
@@ -1,908 +0,0 @@
...
-
-  /** {@code true} if debug indicators should are displayed on images. */
-  @SuppressWarnings("UnusedDeclaration") public boolean getIndicatorsEnabled() {
-    return indicatorsEnabled;
-  }
-
-  /**
-   * Toggle whether debug logging is enabled.
-   * <p>
-   * <b>WARNING:</b> Enabling this will result in excessive object allocation. This should be only
```
Egorand, line 469: Don't see this documentation being migrated, might be useful.

Egorand (approved):

picasso/src/main/java/com/squareup/picasso3/Picasso.java
jrodbx: good catch!

chris-horner (approved):

PR #2274, opened 2022-03-03
Replace checkstyle, errorprone, nullaway with spotless

jrodbx:
...since we're all Kotlin now!

Also, apply formatting suggestions accordingly.


picasso/src/test/java/com/squareup/picasso3/PlatformLruCacheTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011 The Android Open Source Project
+ * Copyright (C) 2011 Square, Inc.
```
Egorand, line 2: Revert? Might need to add an exclude for this file

Egorand (approved):
Nice!

PR #2277, opened 2022-03-24
Fix 3.0 resource loading crash

carranca:
Version 2.8 of Picasso did not require, at hunt completion time, that the request’s uri was non-null (it was always null for resource requests). During Kotlinization we seem to have mistakenly required it to be non-null.

References:
[Picasso 2.8 in Java reading hunter.data.uri without enforcing non-nullability](https://github.com/square/picasso/blob/2.8/picasso/src/main/java/com/squareup/picasso/Picasso.java#L522)
[Picasso 2.8 in Java not requiring non-null values in Listener callback](https://github.com/square/picasso/blob/2.8/picasso/src/main/java/com/squareup/picasso/Picasso.java#L73)


jrodbx:
thanks!

PR #2278, opened 2022-03-25
Remove ViewAnimator from SampleGalleryActivity

gamepro65:
ViewAnimator sets non-visible children to GONE which when used with ImageView causes Picasso's DeferredRequestCreator to never continue with the request because the ImageView never gets laid out.


gamepro65:
Turned out to be related to Don't Keep Activities. When returning to the app after selecting an image the Activity is not laid out before onActivityResult is called which causes this state.

Issue #2275, opened 2022-03-04
Images loaded from the file system render with the wrong orientation


afollestad:
This was originally noticed in POS but I was able to reproduce it in the sample app by using "Load from gallery" and selecting this photo:

<img src="https://user-images.githubusercontent.com/1820165/156857451-351ed204-f4c8-4a51-acdd-9d3df4f9fd56.jpeg" width="300" />

_(If GitHub strips the EXIF data here, `exiftool -Orientation=6 -n [file-name]` could be used to help repro)_

---

It looks like the left image below when loading it using Picasso from a `content://` URI. Seems like `ExifInterface` isn't working here, somehow?

Interestingly, it looks like [BitmapHunter's MatrixTransformation](https://github.com/square/picasso/blob/42adcede6aac7a6ce618f453fdfd3ddaecc43079/picasso/src/main/java/com/squareup/picasso3/BitmapHunter.kt#L151) calling `preRotate` on the `Matrix` is what sets it to the wrong orientation.

<table>
  <tr>
    <th>Current behavior</th>
    <th><a href="https://github.com/square/picasso/blob/42adcede6aac7a6ce618f453fdfd3ddaecc43079/picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.kt#L118">preRotate commented out</a></th>
  </tr>
  <tr>
    <th><img src="https://user-images.githubusercontent.com/1820165/156857527-4edf9c4e-2e4a-44e0-99f6-c1f55362daa9.png" width="400" /></th>
    <th><img src="https://user-images.githubusercontent.com/1820165/156858339-033ad80d-f4c8-4942-8697-ac3e018cc40c.png" width="400" /></th>
  </tr>
</table>

afollestad:
`androidx.exifinterface.media.ExifInterface` seemed to fix it (via a transformation) in POS. Maybe using the AndroidX version of this class in Picasso would solve it?

afollestad:
Interestingly, it seems as though `MatrixTransformation` calling `preRotate` on the Matrix is what sets it to the wrong orientation - and the width/height changes around that.

jrodbx:
Also, part of this is investigating different RequestHandler code paths

PR #2279, opened 2022-03-25
Don't perform EXIf rotation above API 28

gamepro65:
We use the ImageDecoder API on 28+ which is applying the EXIf orientation to the bitmap for us. The alternative to doing a version check here would be to do the version check before setting the exif orientation in each RequestHandler.

Should resolve https://github.com/square/picasso/issues/2275

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.kt
+++ b/picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.kt
@@ -17,6 +17,7 @@ package com.squareup.picasso3
 
 import android.graphics.Bitmap.createBitmap
 import android.graphics.Matrix
+import android.os.Build.VERSION
 import android.view.Gravity
 import androidx.annotation.VisibleForTesting
 import androidx.exifinterface.media.ExifInterface.ORIENTATION_FLIP_HORIZONTAL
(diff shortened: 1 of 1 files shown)


picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.kt
```diff
@@ -111,8 +112,9 @@ internal class MatrixTransformation(private val data: Request) : Transformation
         }
 
         // EXIf interpretation should be done before cropping in case the dimensions need to
-        // be recalculated
-        if (exifOrientation != 0) {
+        // be recalculated; SDK 28+ uses the ImageDecoder API which automatically applies
+        // the EXIf orientation to the bitmap
+        if (exifOrientation != 0 && VERSION.SDK_INT < 28) {
```
carranca, line 117: nit: would this concept be better wrapped in a:

```kotlin
fun shouldPerformExifOrientation() = VERSION.SDK_INT < 28

..

if (exifOrientation != 0 && shouldPerformExifOrientation()) {
```

(as opposed to an inline comment)

carranca:
LGTM but just getting my toes wet on Picasso so would rather not ✅

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.kt
```diff
@@ -227,5 +228,6 @@ internal class MatrixTransformation(private val data: Request) : Transformation
         ORIENTATION_TRANSPOSE, ORIENTATION_TRANSVERSE -> -1
         else -> 1
       }
+
```
jrodbx, line 231: nit: blank line
```suggestion
```

picasso/src/main/java/com/squareup/picasso3/MatrixTransformation.kt
```diff
@@ -111,8 +112,8 @@ internal class MatrixTransformation(private val data: Request) : Transformation
         }
 
         // EXIf interpretation should be done before cropping in case the dimensions need to
-        // be recalculated
-        if (exifOrientation != 0) {
+        // be recalculated; SDK 28+ uses ImageDecoder which handles EXIF orientation
```
jrodbx, line 115: 🎉

jrodbx (approved):

PR #2284, opened 2022-03-30
Minor Test Improvements

luis-cortes:
* Fixed an accidentally duplicated test
* Improved error messaging in a few different tests.

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso3/RequestCreatorTest.kt
+++ b/picasso/src/test/java/com/squareup/picasso3/RequestCreatorTest.kt
@@ -557,7 +557,7 @@ class RequestCreatorTest {
 
   @Test fun invalidCenterInside() {
     try {
-      mockRequestCreator(picasso).resize(10, 10).centerInside().centerCrop()
+      mockRequestCreator(picasso).resize(10, 10).centerCrop().centerInside()
       fail("Calling center inside after center crop should throw exception.")
     } catch (ignored: IllegalStateException) {
     }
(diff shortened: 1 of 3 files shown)


picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.kt
```diff
@@ -165,14 +165,11 @@ class BitmapHunterTest {
     assertThat(eventRecorder.decodedBitmap).isNull()
   }
 
-  @Test fun huntUnrecognizedUri() {
+  @Test(expected = IllegalStateException::class)
```
JakeWharton, line 168: Please never use this. An earlier statement in the body could throw and your test would still pass.

picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.kt
```diff
@@ -971,15 +968,13 @@ class BitmapHunterTest {
     val original = android.graphics.Bitmap.createBitmap(10, 10, ARGB_8888)
     val result = RequestHandler.Result.Bitmap(original, MEMORY, 0)
     val data = Request.Builder(URI_1).build()
-    try {
+    val exception = assertThrows(RuntimeException::class.java) {
```
JakeWharton, line 971: kotlin.test.assertFailsWith reads nicer in Kotlin sources

picasso/src/test/java/com/squareup/picasso3/BitmapHunterTest.kt
luis-cortes: Ah fair enough. Reverting!

picasso/src/test/java/com/squareup/picasso3/RequestCreatorTest.kt
```diff
@@ -557,7 +557,7 @@ class RequestCreatorTest {
 
   @Test fun invalidCenterInside() {
     try {
-      mockRequestCreator(picasso).resize(10, 10).centerInside().centerCrop()
+      mockRequestCreator(picasso).resize(10, 10).centerCrop().centerInside()
```
jrodbx, line 560: oh wow, good eye

jrodbx (approved):

PR #2292, opened 2022-04-13
Experiment with alternate delayed size resolution

chris-horner:
Initially tried integrating with dispatcher but it got hectic. Ended up with essentially a more generic `DeferredRequestCreator` which I'm not sure I like.


picasso/src/main/java/com/squareup/picasso3/RequestCreator.kt
```diff
@@ -433,6 +440,17 @@ class RequestCreator internal constructor(
       return
     }
 
+    if (sizeResolver != null) {
+      picasso.resolveSize(sizeResolver!!, target) { (width, height) ->
```
chris-horner, line 444: Deferred size resolution for `BitmapTarget`

PR #2295, opened 2022-04-28
Add Picasso#rememberPainter extension function under new :picasso-compose module

gamepro65:
Introduces a new :picasso-compose module which exposes a Picasso#rememberPainter function for use with Compose UI. To achieve this Picasso exposes a new .into(DrawableTarget) that way we can provide a PicassoDrawable to the Painter which will keep the API consistent with how we load .into(ImageView).

https://user-images.githubusercontent.com/672316/165820488-19354b63-a588-4cea-bdeb-5186324231dd.mp4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     'targetSdk': 31,
     'sourceCompatibility': JavaVersion.VERSION_1_8,
     'targetCompatibility': JavaVersion.VERSION_1_8,
-    'kotlin': '1.5.21',
+    'kotlin': '1.6.10',
     'okhttp': '4.9.3',
     'okio': '3.0.0',
   ]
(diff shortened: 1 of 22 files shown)


picasso-compose/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=picasso-compose
+POM_NAME=Picasso Compose
+POM_DESCRIPTION=A Compose UI plugin for Picasso.
```
jrodbx, line 3:
```suggestion
POM_DESCRIPTION=Compose UI support for Picasso.
```

jrodbx (approved):
nice!

build.gradle
```diff
@@ -28,6 +28,11 @@ buildscript {
     truth: 'com.google.truth:truth:1.1.3',
     robolectric: 'org.robolectric:robolectric:4.6.1',
     mockito: 'org.mockito:mockito-core:3.0.0',
+    appCompat: 'androidx.appcompat:appcompat:1.3.1',
```
jrodbx, line 31: remove?

build.gradle
```diff
@@ -28,6 +28,11 @@ buildscript {
...
+    drawablePainter: 'com.google.accompanist:accompanist-drawablepainter:0.19.0',
+    composeUi: 'androidx.compose.ui:ui:1.1.1',
+    composeUiTooling: 'androidx.compose.ui:ui-tooling:1.1.1',
```
jrodbx, line 34: remove?

arriolac:
Thanks for providing this—this is great! Is there an issue we can follow to see when the `picasso-compose` artifact will be available on Maven?

PR #2298, opened 2022-05-12
Make `placeholder(Drawable)` parameter nullable in `RequestCreator`.

afollestad:
It was nullable in the previous version of Picasso, and still should be - as the kdoc mentions.


jrodbx (approved):
thanks!

PR #2301, opened 2022-06-23
Port PicassoDrawableTest to instrumented tests

chris-horner:
Adapted [from here](https://github.com/square/picasso/compare/master...jrod/2022-03-30/instrumentation-tests), but ported a less hairy test.


jrodbx (approved):
Mind rebasing on https://github.com/square/picasso/pull/2306 which is slightly different after fixing per comments?

PR #2304, opened 2022-06-28
Add painter optimization to load Picasso once canvas size known

gamepro65:
Adds an "optimizeCanvasSize" property to the painter which delays loading picasso until DrawScope in order to apply an appropriate resize to the RequestCreator based on DrawScope's size. We apply the resize with only one dimension so picasso maintains the aspect ratio otherwise supplying both creates a fill effect. 

https://user-images.githubusercontent.com/672316/176442056-8049ce0f-7ead-4c36-ba66-ec59f5fa7c06.mp4


gamepro65:
Not quite right

PR #2306, opened 2022-06-30
Support instrumentation tests

jrodbx:
Basic CI setup

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -33,11 +33,41 @@ jobs:
           path: |
             **/build/reports/tests/test/
 
+  instrumentation-tests:
+    # We build on a Mac to get hardware acceleration for the Android emulator.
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -33,11 +33,41 @@ jobs:
           path: |
             **/build/reports/tests/test/
 
+  android-emulator:
```
chris-horner, line 36: Would a more descriptive name for this be something like `instrumented-tests`? Seeing `android-emulator` as a job is a bit confusing.

chris-horner (approved):

.github/workflows/build.yaml
jrodbx: Changed to `instrumentation-tests` per LeakCanary: https://github.com/square/leakcanary/blob/main/.github/workflows/main.yml#L28

PR #2313, opened 2022-08-21
Add default Callback implementations.

Isira-Seneviratne:
Add default method implementations for the `Callback` interface.


jrodbx (approved):
thanks!

jrodbx:
might need to add something on the readme for consumers, since I think this now requires them to enable this compiler flag?

PR #2316, opened 2022-09-18
Resolve memory leaks in v3 created during Kotlin conversion

afollestad:
Context: https://square.slack.com/archives/C0366FCTYKC/p1663532908850359

## Commits

* Clear `targetToAction` map in `Picasso.shutdown()`.
* Make a few map properties WeakHashMaps again, to avoid leaks.

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Dispatcher.kt
+++ b/picasso/src/main/java/com/squareup/picasso3/Dispatcher.kt
@@ -51,6 +51,7 @@ import com.squareup.picasso3.Utils.getLogIdsForHunter
 import com.squareup.picasso3.Utils.hasPermission
 import com.squareup.picasso3.Utils.isAirplaneModeOn
 import com.squareup.picasso3.Utils.log
+import java.util.WeakHashMap
 import java.util.concurrent.ExecutorService
 
 internal class Dispatcher internal constructor(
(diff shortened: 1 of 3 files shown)


afollestad:
Should `PicassoTest.shutdown()` be updated to modify the map and assert that it's empty? CC @jrodbx @gamepro65

picasso/src/main/java/com/squareup/picasso3/Picasso.kt
```diff
@@ -378,6 +378,7 @@ class Picasso internal constructor(
     for (deferredRequestCreator in targetToDeferredRequestCreator.values) {
       deferredRequestCreator.cancel()
     }
+    targetToAction.clear()
```
jrodbx, line 381: note to self: according to Aidan, this is only happening as of v3.  so is this a kotlin conversion bug or just something we never caught in v2?

jrodbx (approved):

jrodbx:
> Should `PicassoTest.shutdown()` be updated to modify the map and assert that it's empty?

Maybe add a new shutdown variant test to test that case, but yes to having a test 👍

picasso/src/main/java/com/squareup/picasso3/Dispatcher.kt
```diff
@@ -63,10 +64,10 @@ internal class Dispatcher internal constructor(
   internal val hunterMap = mutableMapOf<String, BitmapHunter>()
 
   @get:JvmName("-failedActions")
-  internal val failedActions = mutableMapOf<Any, Action>()
+  internal val failedActions = WeakHashMap<Any, Action>()
 
   @get:JvmName("-pausedActions")
-  internal val pausedActions = mutableMapOf<Any, Action>()
+  internal val pausedActions = WeakHashMap<Any, Action>()
```
jrodbx, line 70: This does result in using the JVM types again, which was probably the thing we were mistakenly optimizing for in the conversion.

jrodbx (approved):

picasso/src/main/java/com/squareup/picasso3/Dispatcher.kt
afollestad: Yeah 😞 I wish Kotlin had a "first-party" equivalent

jrodbx:
please rebase onto latest now that https://github.com/square/picasso/pull/2318 has landed

afollestad:
> please rebase onto latest now that #2318 has landed

@jrodbx looks like that PR is still open 😅

afollestad:
Looks like I don't have merge permission 🙂

PR #2317, opened 2022-09-20
Bump compose 1.2.0

gamepro65:


jrodbx:
please rebase onto latest now that https://github.com/square/picasso/pull/2318 has landed

PR #2322, opened 2022-09-22
Update `Action` to store targets with `WeakReference`.

afollestad:


picasso/src/main/java/com/squareup/picasso3/Action.kt
```diff
@@ -16,19 +16,20 @@
...
 
-internal abstract class Action(
+internal abstract class Action<T>(
   val picasso: Picasso,
   val request: Request
 ) {
   var willReplay = false
   var cancelled = false
 
+  abstract val target: WeakReference<T>
```
pyricau, line 28: Consider going one step further and making WeakReference an implementation detail of target, and target final.

Something like this:

```kotlin
internal abstract class Action<T>(
  val picasso: Picasso,
  val request: Request
  target: T
) {

  private val targetRef = WeakReference(target)
  
  val target: T?
    get() = targetRef.get()

}

```

pyricau:
LGTM though I think target could be nicer maybe (see suggestion). Not "approving" the PR as it's not my call ;) .

picasso/src/main/java/com/squareup/picasso3/Action.kt
afollestad: Good idea! Thanks @pyricau 🙏🏻

afollestad: Though if I go this route, `FetchAction` and `GetAction` get a little weird because they don't necessarily want callers to provide a target

afollestad: @jrodbx &

afollestad: Resolved!

jrodbx:
We did some refactoring to move the target out of the abstract Action class and into the specific subtypes.  See https://github.com/square/picasso/commit/e4f5b3c7d96b8c29ca1663aca5dbac913d4ba031 and https://github.com/square/picasso/commit/5e7659afe6512f1d20a69a5e6ba069c3e74ed363.

In the former, notice that the T target being passed to super() is null for FetchAction, GetAction, RemoteViewsAction (and therefore its subtypes), and non-null for ImageViewAction and BitmapTargetAction (formerly TargetAction).

So I think we want to keep the WeakReferences to those two Actions only.  How about something like this?  https://github.com/square/picasso/pull/2329

PR #2324, opened 2022-09-27
Use NotificationChannelCompat.

Isira-Seneviratne:
* Use `NotificationChannelCompat` in the sample app.
* Fix crash on API 31+ when selecting the "Sample Notification" option.


jrodbx (approved):
thanks!

PR #2325, opened 2022-09-27
Remove BitmapCompat.getAllocationByteCount().

Isira-Seneviratne:
Remove calls to `BitmapCompat.getAllocationByteCount()`, as the minimum API level is 21 and [`Bitmap#getAllocationByteCount()`](https://developer.android.com/reference/android/graphics/Bitmap#getAllocationByteCount()) is supported from API level 19 onward.


jrodbx (approved):
thanks!

PR #2327, opened 2022-09-28
Use AndroidX's LruCache.

Isira-Seneviratne:
Use the `LruCache` class in AndroidX Collection.


jrodbx:
mind rebasing?

picasso/src/main/java/com/squareup/picasso3/PlatformLruCache.kt
```diff
@@ -16,20 +16,15 @@
...
-  val cache =
-    object : LruCache<String, BitmapAndSize>(if (maxByteCount != 0) maxByteCount else 1) {
-      override fun sizeOf(
-        key: String,
-        value: BitmapAndSize
-      ): Int = value.byteCount
-    }
+  val cache = lruCache<String, BitmapAndSize>(
+    maxByteCount.coerceAtLeast(1), { _, value -> value.byteCount }
+  )
```
afollestad, line 27:
```suggestion
  val cache = lruCache<String, BitmapAndSize>(
    maxByteCount.coerceAtLeast(1)
  ) { _, value -> value.byteCount }
```
The lambda can be pulled out of the parantheses

Isira-Seneviratne:
> mind rebasing?

Sure, no problem.

picasso/src/main/java/com/squareup/picasso3/PlatformLruCache.kt
Isira-Seneviratne: The final lambda parameter is for the `entryRemoved` method, so it gives a compiler error.

PR #2332, opened 2022-10-20
Integrate LeakCanary for instrumentation tests, and the sample app.

afollestad:
Not sure if it's even worth enabling in _current_ instrumentation tests. But it's there 😄


afollestad:
Shout out to @py for great documentation and a pretty effortless setup process!

afollestad:
Shout out to @pyricau for great documentation and a pretty effortless setup process!

pyricau:
Looks good, my only question is, do these test involve any kind of thing that has a lifecycle like activity, fragment or views? If not, then leakcanary's not going to get triggered, we need to give it something (e.g. a bitmap) for it to detect if the thing is leaking.

afollestad:
> Not sure if it's even worth enabling in _current_ instrumentation tests.

Not really, which is why I said "Not sure if it's even worth enabling in current instrumentation tests" 😄

PR #2333, opened 2022-10-20
Revert "Remove flushing our dispatcher"

gamepro65:
Unfortunately it seems this bug didn't get fixed in Android 5 but did eventually get fixed in Android 12. Updated the comment to reflect LeakCanary's [tracking](https://github.com/square/leakcanary/blob/main/plumber-android-core/src/main/java/leakcanary/AndroidLeakFixes.kt#L153) of the bug


jrodbx (approved):
Bummer

PR #2341, opened 2022-11-09
Prevent PicassoExecutorService race condition in DispatcherTest

gamepro65:
Been seeing a pattern of flaky DispatcherTests along the lines of

```
com.squareup.picasso3.DispatcherTest > performSubmitWithTwoDifferentRequestsQueuesHunters FAILED
    value of: map.size()
    expected: 2
    but was : 1
    map was : {}
        at com.squareup.picasso3.DispatcherTest.performSubmitWithTwoDifferentRequestsQueuesHunters(DispatcherTest.kt:110)
```

With the debugger I was able to see that PicassoExecutorService is actually spinning up a new thread and executing / completing before our test asserts causing the failure / race condition. Tried swapping PicassoExecutorService out but its required for some tests, so using a spy to prevent creating / executing other threads.

### DIFF
--- a/picasso/src/test/java/com/squareup/picasso3/DispatcherTest.kt
+++ b/picasso/src/test/java/com/squareup/picasso3/DispatcherTest.kt
@@ -38,6 +38,7 @@ import com.squareup.picasso3.TestUtils.URI_1
 import com.squareup.picasso3.TestUtils.URI_2
 import com.squareup.picasso3.TestUtils.URI_KEY_1
 import com.squareup.picasso3.TestUtils.URI_KEY_2
+import com.squareup.picasso3.TestUtils.any
 import com.squareup.picasso3.TestUtils.makeBitmap
 import com.squareup.picasso3.TestUtils.mockAction
 import com.squareup.picasso3.TestUtils.mockBitmapTarget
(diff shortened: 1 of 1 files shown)


picasso/src/test/java/com/squareup/picasso3/DispatcherTest.kt
```diff
@@ -71,13 +75,15 @@ class DispatcherTest {
...
+  private val executorService = spy(PicassoExecutorService())
   private val cache = PlatformLruCache(2048)
-  private val service = TestDelegatingService(PicassoExecutorService())
+  private val service = TestDelegatingService(executorService)
   private val bitmap1 = makeBitmap()
 
   @Before fun setUp() {
     initMocks(this)
     `when`(context.applicationContext).thenReturn(context)
+    doReturn(mock(Future::class.java)).`when`(executorService).submit(any(Runnable::class.java))
```
afollestad, line 86: Do you guys not bring in Mockito Kotlin?

afollestad (approved):

picasso/src/test/java/com/squareup/picasso3/DispatcherTest.kt
jrodbx: not at the moment, cuz we'd like to remove all mocks, but that will take a bit of refactoring to achieve.

jrodbx (approved):

PR #2342, opened 2022-11-28
Add gif support to network requests

gamepro65:
Uses ImageDecoder when available which already handles gifs and falls back to using the `android.graphics.Movie` API when < 28 with a custom Drawable for rendering. 

https://user-images.githubusercontent.com/672316/204376251-37543b2a-b9d7-499c-ba16-96bfc4853aaf.mov


picasso/src/main/java/com/squareup/picasso3/MovieDrawable.kt
```diff
@@ -0,0 +1,70 @@
...
+import android.graphics.Canvas
+import android.graphics.ColorFilter
+import android.graphics.Movie
+import android.graphics.Paint
+import android.graphics.PixelFormat
+import android.graphics.drawable.Animatable
+import android.graphics.drawable.Drawable
+import android.os.SystemClock
+
+internal class MovieDrawable(private val movie: Movie) : Drawable(), Animatable {
```
afollestad, line 27: 🤯

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.kt
```diff
@@ -67,8 +68,14 @@ internal class NetworkRequestHandler(
...
-            val bitmap = decodeStream(body!!.source(), request)
-            callback.onSuccess(Result.Bitmap(bitmap, loadedFrom))
+            val bufferedSource = body!!.source()
+            if (Utils.isGifFile(bufferedSource)) {
+              val drawable = decodeDrawableStream(bufferedSource, request)
+              callback.onSuccess(Result.Drawable(drawable, loadedFrom))
+            } else {
+              val bitmap = decodeStream(bufferedSource, request)
+              callback.onSuccess(Result.Bitmap(bitmap, loadedFrom))
+            }
```
afollestad, line 78: Do you think it would be more readable like this?

```suggestion
            val result = if (Utils.isGifFile(bufferedSource)) {
              val drawable = decodeDrawableStream(bufferedSource, request)
              Result.Drawable(drawable, loadedFrom)
            } else {
              val bitmap = decodeStream(bufferedSource, request)
              Result.Bitmap(bitmap, loadedFrom)
            }
            callback.onSuccess(result)
```

Also, can `isGifFile` be static imported here?

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.kt
```diff
@@ -186,6 +189,46 @@ internal object BitmapUtils {
...
+   */
+  fun decodeDrawableStream(source: Source, request: Request): Drawable {
+    val exceptionCatchingSource = ExceptionCatchingSource(source)
+    val bufferedSource = exceptionCatchingSource.buffer()
+    val drawable =
+      if (VERSION.SDK_INT >= 28) {
+        val imageSource = ImageDecoder.createSource(ByteBuffer.wrap(bufferedSource.readByteArray()))
+        decodeDrawableSourceP(imageSource, request)
+      } else {
+        if (Utils.isGifFile(bufferedSource)) {
```
afollestad, line 204: Can we static import `isGifFile`?

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.kt
```diff
@@ -186,6 +189,46 @@ internal object BitmapUtils {
...
+  /**
+   * Decode a byte stream into a Drawable. This method will take into account additional information
+   * about the supplied request in order to do the decoding efficiently.
+   */
+  fun decodeDrawableStream(source: Source, request: Request): Drawable {
+    val exceptionCatchingSource = ExceptionCatchingSource(source)
+    val bufferedSource = exceptionCatchingSource.buffer()
+    val drawable =
+      if (VERSION.SDK_INT >= 28) {
+        val imageSource = ImageDecoder.createSource(ByteBuffer.wrap(bufferedSource.readByteArray()))
```
afollestad, line 201: I feel like in POS, we like to extract many things out to extension functions to improve readability, but maybe this would be an overkill here?

```kotlin
private fun BufferedSource.toImageDecoder(): ImageDecoder.Source =
  ImageDecoder.createSource(wrap(readByteArray()))
```

afollestad (approved):
So cool

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.kt
afollestad: With that line reduced the `if` statement might be able to be pulled up next to the `drawable` declaration?

picasso/src/main/java/com/squareup/picasso3/NetworkRequestHandler.kt
gamepro65: 👍 I like extracting the result and don't have a strong preference with the import. Looking around the code base it seems mixed with doing the full static import

picasso/src/main/java/com/squareup/picasso3/BitmapUtils.kt
gamepro65: I think the extension function is overkill but am able to pull the if up a line which does help a bit

haresh-simform (approved):
Amazing i can use this approach in react native side also with bridging great

PR #2350, opened 2023-01-26
Upgrade Gradle version to 7.6

sconvent:
This will upgrade the Gradle version defined in the Gradle wrapper to 7.6.

Done using `./gradlew wrapper --gradle-version 7.6 --distribution-type all`


jrodbx (approved):
thanks!

PR #2423, opened 2023-09-01
Proof-of-concept: Make PicassoPainter's request restart when read state is changed

zach-klippenstein:
Allows `rememberPainter` to support the use case where the `request` function reads some snapshot state that gets changed between the initial `rememberPainter` call and the first `onDraw` call. For example:
```kotlin
@Composable fun MyImage() {
  var size: IntSize by remember { mutableStateOf(IntSize.Zero) }
  Image(
    modifier = Modifier.onSizeChanged { size = it },
    painter = picasso.rememberPainter {
      // 1. Composition: This will get called immediately, before size has
      //    been initialized by the first layout phase.
      //        - size=IntSize.Zero
      //        - load(null)
      //        - painter=placeholder
      // 2. Layout: onSizeChanged invoked and size is updated.
      //        - size=IntSize(width, height)
      //        - load not called yet
      //        - painter=placeholder (unchanged)
      // 3. Draw: Before Painter draws loaded image, it will see the
      //    dependencies have changed and call this lambda again. If the
      //    resulting image request can be loaded from memory, it will be
      //    immediately available for drawing in that same frame.
      //        - size=IntSize(width, height)
      //        - load("some://url.with/size?width,height")
      //        - if image was cached in memory,
      //            - painter=loaded image
      //            - painter=placeholder until async load finishes
      val imageUrl = if (size == IntSize.Zero) null else getImageUrl(size = size)
      it.load(imageUrl)
    }
  )
}
```

This change ensures that if `request` has any state dependencies, that when those dependencies are changed:
1. Composition is _not_ invalidated.
    - Previously, since `request` was called from `onRemembered`, the first change would invalidate the composition, and then subsequent changes wouldn't invalidate anything.
2. Anything that read the `PicassoPainter`'s `intrinsicSize` or called `draw` _is_ invalidated.

It also ensures that any time `intrinsicSize` or `onDraw` is called, we call `RequestCreator.into` if the request has been invalidated, before consuming the drawable, to allow the latest `RequestCreator` to be loaded from memory if available. It relies on `derivedStateOf` returning the cached instance to avoid executing redundant `into` calls if `request`'s dependencies haven't changed.

This approach unfortunately means that the `request` function might run during the draw phase, which means that at the very least a new `RequestCreator` would be allocated, in addition to whatever other work the function might be doing. It's generally not good to allocate or do any non-draw-related work during the draw pass, but I don't think it can be avoided. Because `Painter`s don't necessarily participate in layout, `onDraw` is the only hook we have that's guaranteed to run after any layout code (as required in the example above) but still in the current frame. If Picasso's public Compose API was something that participated in layout directly, like an `Image`-like composable or a modifier, then this could be done more efficiently since the `request` function would only need to be checked before its own measurement, not in every draw pass.

### DIFF
--- a/picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
+++ b/picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
@@ -40,31 +40,14 @@ fun Picasso.rememberPainter(
   request: (Picasso) -> RequestCreator
 ): Painter {
   return remember(key) { PicassoPainter(this, request, onError) }
-    .also { painter ->
-      DisposableEffect(painter) {
-        painter.start()
-        onDispose(painter::stop)
-      }
-    }
 }
 
 internal class PicassoPainter(
   private val picasso: Picasso,
-  request: (Picasso) -> RequestCreator,
+  private val request: (Picasso) -> RequestCreator,
   private val onError: ((Exception) -> Unit)? = null
 ) : Painter(), RememberObserver, DrawableTarget {
 
-  private val stateObserver = SnapshotStateObserver(onChangeExecutor = { callback ->
-      if (Looper.myLooper == Looper.mainLooper) {
-        callback()
-      } else {
-        postToMainThread(callback)
-      }
-    })
-  private val request: RequestCreator by derivedStateOf { request.invoke(picasso) }
-  // This is technically written to from composition once, by onRemembered. However it's
-  // never written to from composition again so it doesn't need to be snapshot state.
-  private var lastRequest: RequestCreator? = null
   private var painter: Painter by mutableStateOf(EmptyPainter)
   private var alpha: Float by mutableStateOf(DefaultAlpha)
   private var colorFilter: ColorFilter? by mutableStateOf(null)
@@ -89,7 +72,7 @@ internal class PicassoPainter(
   }
 
   override fun onRemembered() {
-    loadRequest()
+    request.invoke(picasso).into(this)
   }
 
   override fun onAbandoned() {
@@ -102,39 +85,6 @@ internal class PicassoPainter(
     (painter as? RememberObserver)?.onForgotten()
     painter = EmptyPainter
     picasso.cancelRequest(this)
-  }
-
-  fun start() {
-    stateObserver.start()
-    // Even though loadRequest was called in onRemembered, we need to call it
-    // again now that the observer is started to observe state reads.
-    loadRequest()
-  }
-
-  fun stop() {
-    stateObserver.stop()
-    stateObserver.clear()
-    // Picasso state will be cleaned up by onForgotten, we don't need to
-    // do it here.
-  }
-
-  private fun loadRequest() {
-    // Because this will get called from composition, and we don't care about
-    // lastRequest changing and are handling request reads ourself.
-    // It will get called again from start() and that will observe the reads.
-    Snapshot.withoutReadObservation {
-      // The first time this is called, from composition, the observer won't
-      // have been started yet so this will run the block but not observe it.
-      val newRequest = stateObserver.observeReads(this, onCommitAffectingRequest) {
-        request
-      }
-      // Can be != if RequestCreators are comparable. Identity comparison works
-      // because derivedStateOf will return a cached instance if nothing changed.
-      if (newRequest !== lastRequest) {
-        lastRequest = newRequest
-        newRequest.into(this)
-      }
-    }
   }
 
   override fun onPrepareLoad(placeHolderDrawable: Drawable?) {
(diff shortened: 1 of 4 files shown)


picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -40,14 +40,21 @@ fun Picasso.rememberPainter(
...
 }
 
 internal class PicassoPainter(
   private val picasso: Picasso,
   private val request: (Picasso) -> RequestCreator,
   private val onError: ((Exception) -> Unit)? = null
-) : Painter(), RememberObserver, DrawableTarget {
+) : Painter(), DrawableTarget {
 
+  private val stateObserver = SnapshotStateObserver(onChangeExecutor = { it() })
```
zach-klippenstein, line 57: This determines where subsequent calls to `request` will occur. Might want to make this post to the main thread if not already on it.

picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -71,20 +78,23 @@ internal class PicassoPainter(
...
     picasso.cancelRequest(this)
   }
 
-  override fun onForgotten() {
-    (painter as? RememberObserver)?.onForgotten()
-    painter = EmptyPainter
-    picasso.cancelRequest(this)
+  private fun loadRequest() {
+    stateObserver.observeReads(this, onCommitAffectingRequest) {
+      request.invoke(picasso).into(this)
```
zach-klippenstein, line 96: Technically you only need to call `request.invoke(picasso)` inside the `observeReads` lambda. However, since we know `into` isn't doing any state reads we don't care about, it doesn't really matter.

If `RequestCreator`s are comparable, you might want to wrap the `request.invoke(picasso)` call in a `derivedStateOf` stored in a class property to avoid making new requests for equivalent `RequestCreator`s.

picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -40,14 +40,21 @@ fun Picasso.rememberPainter(
   request: (Picasso) -> RequestCreator
 ): Painter {
   return remember(key) { PicassoPainter(this, request, onError) }
```
zach-klippenstein, line 42: It's not strictly related to this change but there's another bug here: `request` and `onError` should either be passed as keys to `remember`, stored in `rememberUpdatedState`, or updated on the painter on recomposition. E.g. if this recomposes while the load is in progress and the caller passes in a new `onError` lambda, the new handler will be ignored.

picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -71,20 +78,23 @@ internal class PicassoPainter(
...
   }
 
-  override fun onForgotten() {
-    (painter as? RememberObserver)?.onForgotten()
-    painter = EmptyPainter
-    picasso.cancelRequest(this)
+  private fun loadRequest() {
+    stateObserver.observeReads(this, onCommitAffectingRequest) {
+      request.invoke(picasso).into(this)
+    }
```
zach-klippenstein, line 97:
```suggestion
    // add `private val previousRequest: RequestCreator? = null` to class 
    val request = stateObserver.observeReads(this, onCommitAffectingRequest) {
      request.invoke(picasso)
    }
    if (request != previousRequest) {
      previousRequest = request
      request.into(this)
    }
```

picasso/src/androidTest/java/com/squareup/picasso3/PicassoComposePainterTest.kt
```diff
@@ -0,0 +1,148 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.google.common.truth.Truth.assertThat
+import com.squareup.picasso3.compose.rememberPainter
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class PicassoComposePainterTest {
```
zach-klippenstein, line 41: Couldn't run these tests because gradle needed more tweaking.

picasso/src/androidTest/java/com/squareup/picasso3/PicassoComposePainterTest.kt
```diff
@@ -0,0 +1,148 @@
...
+                            )
+                        )
+                    }
+
+                    with(painter) {
+                        draw(canvasSize)
+                    }
+
+                    // Draw triggers request was made with size.
+                    assertThat(lastRequest.headers?.toMultimap()).containsAtLeastEntriesIn(
```
zach-klippenstein, line 91: I don't think this will actually pass right now because it's not going to be ran synchronously? Might need to warm the memory cache with the fake request?

PR #2424, opened 2023-09-03
Removed the deprecated OnLifeCycleEvents Annotation & implemented Def…

GibsonRuitiari:
Removed the deprecated OnLifeCycleEvents Annotation which relied on reflection, and instead made Picasso class to implement the recommended DefaultLifecycleObserver .


jrodbx (approved):
Thanks!  

~Please run `./gradlew picasso:apiDump` and commit the results.~. nvm, just added a commit

PR #2426, opened 2023-09-05
Fix Request's newBuilder not copying headers

gamepro65:
We weren't copying the headers 😢

### DIFF
--- a/picasso/src/main/java/com/squareup/picasso3/Request.kt
+++ b/picasso/src/main/java/com/squareup/picasso3/Request.kt
@@ -330,6 +330,7 @@ class Request internal constructor(builder: Builder) {
       priority = request.priority
       memoryPolicy = request.memoryPolicy
       networkPolicy = request.networkPolicy
+      headers = request.headers
     }
 
     fun hasImage(): Boolean {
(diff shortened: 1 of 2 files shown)


picasso/src/main/java/com/squareup/picasso3/Request.kt
```diff
@@ -330,6 +337,7 @@ class Request internal constructor(builder: Builder) {
       priority = request.priority
       memoryPolicy = request.memoryPolicy
       networkPolicy = request.networkPolicy
+      headers = request.headers
```
gamepro65, line 340: Headers were also not being copied when we utilize newBuilder() like in BitmapHunter

```
    if (retryCount == 0) {
      data = data.newBuilder().networkPolicy(NetworkPolicy.OFFLINE).build()
    }
```

picasso/src/main/java/com/squareup/picasso3/Request.kt
```diff
@@ -214,6 +214,13 @@ class Request internal constructor(builder: Builder) {
       val path = data.uri.toString()
       builder.ensureCapacity(path.length + KEY_PADDING)
       builder.append(path)
+
+      data.headers?.toMultimap()?.let {
+        builder
+          .append(KEY_SEPARATOR)
+          .append("headers:")
+          .append(it)
+      }
```
jrodbx, line 223: i'm unsure about this.  if say an OAuth token gets refreshed for an authenticated endpoint, a subsequent image request might result in an unnecessary cache miss.

picasso/src/main/java/com/squareup/picasso3/Request.kt
jrodbx: good catch!

jrodbx (approved):

PR #2432, opened 2023-09-11
Make PicassoPainter's request aware of state read changes

gamepro65:
Based on https://github.com/square/picasso/pull/2423

This allows state objects the ability to update a request without requiring a recomposition + another rememberPainter call with a different key. If a state change occurs during the initial composition (ex. as a result of measured size becoming available) then we need to cancel the original request. Unfortunately Picasso will already be executing the original request; to avoid this race condition I've included a main synchronization mechanism to prevent the request from running until the current frame has completed which allows us time to cancel the initial request before processing anything unnecessarily. This only effects cases where the image does not exist in the cache already as those would be synchronously returned in time for the first frame.

### DIFF
--- a/picasso-compose/src/androidTest/java/com/squareup/picasso3/compose/PicassoPainterTest.kt
+++ b/picasso-compose/src/androidTest/java/com/squareup/picasso3/compose/PicassoPainterTest.kt
@@ -1,0 +1,197 @@
+/*
+ * Copyright (C) 2013 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.picasso3.compose
+
+import android.graphics.Bitmap
+import android.graphics.Bitmap.Config.ARGB_8888
+import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.requiredSize
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.onSizeChanged
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.IntSize
+import androidx.compose.ui.unit.dp
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.google.common.truth.Truth.assertThat
+import com.squareup.picasso3.Picasso
+import com.squareup.picasso3.Picasso.LoadedFrom
+import com.squareup.picasso3.Request
+import com.squareup.picasso3.RequestHandler
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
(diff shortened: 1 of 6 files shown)


picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -48,12 +51,19 @@ internal class PicassoPainter(
   private val onError: ((Exception) -> Unit)? = null
 ) : Painter(), RememberObserver, DrawableTarget {
 
+  private var lastRequestCreator: RequestCreator? by mutableStateOf(null)
+  private val requestCreator: State<RequestCreator> = derivedStateOf { request(picasso) }
```
zach-klippenstein, line 55: Nit: Just for consistency with the other states here, this could use property delegation as well.

picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -100,6 +117,20 @@ internal class PicassoPainter(
...
 
+  private fun load() {
+    // This derived state read will return the same instance of RequestCreator if one has been
+    // cached and none of the state dependencies have since changed.
+    val requestCreator = requestCreator.value
+    // lastRequestCreator is just used for diffing, we don't want it to invalidate anything.
+    val lastRequestCreator = Snapshot.withoutReadObservation { lastRequestCreator }
+
+    // Only launch a new request if anything has actually changed.
+    if (requestCreator != lastRequestCreator) {
```
zach-klippenstein, line 128: I think `RequestCreator` doesn't implement `equals` so this is effectively just a reference check. Might be worth implementing `equals` to avoid even more re-requests. Might be worth implementing `equals` to make this more meaningful. Or at least documenting that limitation with a comment here.

zach-klippenstein (approved):
Compose stuff LGTM.

picasso-compose/src/androidTest/java/com/squareup/picasso3/compose/PicassoPainterTest.kt
```diff
@@ -0,0 +1,197 @@
...
+      .build()
+
+    var drawInvalidator by mutableStateOf(0)
+    var drawCount = 0
+    rule.setContent {
+      CompositionLocalProvider(LocalDensity provides Density(1f)) {
+        val painter = picasso.rememberPainter {
+          it.load("http://example.com/")
+        }
+        Canvas(Modifier.fillMaxSize()) {
```
jrodbx, line 129: would it be helpful to use a named argument for the lambda here?

picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -66,13 +75,20 @@ internal class PicassoPainter(
...
     }
   }
 
   override fun onRemembered() {
-    request.invoke(picasso).into(this)
+    // This is called from composition, but if the request provider function reads any state we
+    // don't want that to invalidate composition. It will invalidate draw, later.
+    Snapshot.withoutReadObservation {
+      load()
+    }
```
jrodbx, line 91:
```suggestion
    Snapshot.withoutReadObservation { load() }
```

nit

picasso-compose/src/main/java/com/squareup/picasso3/compose/PicassoPainter.kt
```diff
@@ -100,6 +116,22 @@ internal class PicassoPainter(
...
+  private fun load() {
+    // This derived state read will return the same instance of RequestCreator if one has been
+    // cached and none of the state dependencies have since changed.
+    val requestCreator = requestCreator
+    // lastRequestCreator is just used for diffing, we don't want it to invalidate anything.
+    val lastRequestCreator = Snapshot.withoutReadObservation { lastRequestCreator }
+
+    // Only launch a new request if anything has actually changed. RequestCreator does not
+    // currently implement an equals method, relying here on reference equality, future improvement
+    // will be to implement equals which can prevent further re-requests.
```
jrodbx, line 128: nice comments/explanations

jrodbx (approved):
awesome, thanks for improving this!

picasso-compose/src/androidTest/java/com/squareup/picasso3/compose/PicassoPainterTest.kt
gamepro65: I don't think so, its the only other argument and generally accepted that Canvas' lambda is DrawScope

Issue #2472, opened 2024-04-29
shutdown() throws ClosedReceiveChannelException when using coroutines dispatcher


jingibus:
The fix is straightforward. Run the job in `InternalCoroutineDispatcher` like so:

```kotlin
  init {
    // Using a channel to enforce sequential access for this class' internal state
    scope.launch {
      channel.receiveAsFlow().collect {
        it.invoke()
      }
    }
  }
```

A repro isn't trivial, though. For it to happen, `channel.close()` has to actually wake up a suspended coroutine. This happens consistently for me in a live app, but I can't see how to make it happen in unit tests with the current structure of the dispatcher. I think we would need `performSubmit` to actually block and do some work to exercise this scenario.

jingibus:
I think that fix will work, at least. It doesn't make a lot of sense to me that `close()` should throw like that, but it does make a lot of sense to me that code with a bare `receive().invoke()` like in the current code should throw on close:code:

```kotlin
  init {
    // Using a channel to enforce sequential access for this class' internal state
    scope.launch {
      while (!channel.isClosedForReceive) {
        channel.receive().invoke()
      }
    }
  }
```

 I believe `close()` is just bubbling up that exception to that call site in an odd way.

PR #2473, opened 2024-05-01
Prevent ClosedReceiveChannelException after shutdown

gamepro65:
The channel is always suspended at `channel.receive()` meaning anything called on the channel after shutdown would still wake up this suspended call-site and crash because we are closed.

Resolves: https://github.com/square/picasso/issues/2472


jrodbx (approved):
Nice test!

